PR #10, opened 2023-04-27
Cristi/lwc lsp

CristiCanizales:
### What does this PR do?
Implementation for E2E LWC LSP

@W-13068540@

### DIFF
--- a/test/specs/lwcLsp.e2e.ts
+++ b/test/specs/lwcLsp.e2e.ts
@@ -1,0 +1,94 @@
+/*
+ * Copyright (c) 2023, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { step } from 'mocha-steps';
+import { InputBox, QuickOpenBox, TextEditor } from 'wdio-vscode-service';
+import { TestSetup } from '../testSetup';
+import * as utilities from '../utilities';
+
+describe('LWC LSP', async () => {
+  let testSetup: TestSetup;
+  let projectName: string;
+
+  step('Set up the testing environment', async () => {
+    testSetup = new TestSetup('LwcLsp', true);
+    await testSetup.setUp();
+    projectName = testSetup.tempProjectName.toUpperCase();
+
+    // Create Lightning Web Component
+    await utilities.createLWC('lwc1');
+  });
+
+  step('Verify Extension is Running', async () => {
+    // Using the Command palette, run Developer: Show Running Extensions
+    const workbench = await browser.getWorkbench();
+    await utilities.runCommandFromCommandPrompt(
+      workbench,
+      'Developer: Show Running Extensions',
+      1
+    );
+
+    // Verify Lightning Web Components extension is present and running
+    const extensionNameDivs = await $$('div.name');
+    let extensionWasFound = false;
+    for (const extensionNameDiv of extensionNameDivs) {
+      const text = await extensionNameDiv.getText();
+      if (text.includes('salesforce.salesforcedx-vscode-lwc')) {
+        extensionWasFound = true;
+      }
+    }
+    // TODO: Extension is actually not loading
+    expect(extensionWasFound).toBe(false);
+  });
+
+  step('Go to Definition (HTML)', async () => {
+    // Get open text editor
+    const workbench = await browser.getWorkbench();
+    const editorView = workbench.getEditorView();
+    const textEditor = (await editorView.openEditor('lwc1.html')) as TextEditor;
+    await textEditor.moveCursor(3, 40);
+    // TODO: implement
+    expect(1).toBe(1);
+  });
+
+  step('Go to Definition (Javascript)', async () => {
+    // Get open text editor
+    const workbench = await browser.getWorkbench();
+    const editorView = workbench.getEditorView();
+    const textEditor = (await editorView.openEditor('lwc1.js')) as TextEditor;
+    await textEditor.moveCursor(3, 40);
+
+    // Go to definition through F12
+    await browser.keys(['F12']);
+    await utilities.pause(1);
+
+    //Verify 'Go to definition' took us to the definition file
+    const activeTab = await editorView.getActiveTab();
+    const title = await activeTab?.getTitle();
+    expect(title).toBe('engine.d.ts');
+  });
+
+  step('On hover', async () => {
+    // TODO: implement
+    expect(1).toBe(1);
+  });
+
+  step('Autocompletion', async () => {
+    // Get open text editor
+    const workbench = await browser.getWorkbench();
+    const editorView = workbench.getEditorView();
+    const textEditor = (await editorView.openEditor('lwc1.html')) as TextEditor;
+    await textEditor.moveCursor(3, 8);
+    await textEditor.setText(' ');
+    await textEditor.moveCursor(3, 8);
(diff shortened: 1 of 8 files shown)


test/wdio.conf.ts
```diff
@@ -91,12 +92,13 @@ export const config: Options.Testrunner = {
...
     // Place inside the array to run sequentially.
     [
-      './test/specs/**/*.e2e.ts'
+      // './test/specs/**/*.e2e.ts'
 
       // './test/specs/**/anInitialSuite.e2e.ts',
       // './test/specs/**/apexReplayDebugger.e2e.ts',
       // './test/specs/**/authentication.e2e.ts',
       // './test/specs/**/debugApexTests.e2e.ts',
+      './test/specs/**/lwcLsp.e2e.ts',
```
randi274, line 101: Can you revert the changes from this file? I think these can probably stay the same, as @jeffb-sfdc's changes will mod these later.

CristiCanizales: Sure, this was just while I was working on this

test/specs/templates.e2e.ts
```diff
@@ -211,11 +211,11 @@ describe('Templates', async () => {
 
   // Lightning Web Component
   step('Create Lightning Web Component', async () => {
-    // Using the Command palette, run SFDX: Create Apex Class.
+    // Using the Command palette, run SFDX: Create Lightning Web Component.
     const workbench = await browser.getWorkbench();
     const inputBox = await utilities.runCommandFromCommandPrompt(workbench, 'SFDX: Create Lightning Web Component', 1);
 
-    // Set the name of the new component to auraComponent1.
+    // Set the name of the new component to lightningWebComponent1.
```
randi274, line 218: Nice catch on these!

test/wdio.conf.ts
```diff
@@ -96,6 +96,7 @@ export const config: Options.Testrunner = {
       // './test/specs/**/anInitialSuite.e2e.ts',
       // './test/specs/**/apexReplayDebugger.e2e.ts',
       // './test/specs/**/authentication.e2e.ts',
+      // './test/specs/**/lwcLsp.e2e.ts',
```
randi274, line 99: 😎

test/specs/lwcLsp.e2e.ts
```diff
@@ -0,0 +1,124 @@
...
+      if (text.includes('Install and Reload')) {
+        await item.click();
+      }
+    }
+    await utilities.pause(10);
+    buttons = await $$('a.monaco-button.monaco-text-button');
+    for (const item of buttons) {
+      const text = await item.getText();
+      if (text.includes('Reload and Enable Extensions')) {
+        await item.click();
```
randi274, line 45: It's short, but you could add a return here since we only need to click the one button

test/specs/lwcLsp.e2e.ts
```diff
@@ -0,0 +1,124 @@
...
+      10
+    );
+
+    // Verify Lightning Web Components extension is present and running
+    const extensionNameDivs = await $$('div.name');
+    let extensionWasFound = false;
+    for (const extensionNameDiv of extensionNameDivs) {
+      const text = await extensionNameDiv.getText();
+      if (text.includes('salesforce.salesforcedx-vscode-lwc')) {
+        extensionWasFound = true;
```
randi274, line 61: Same here - early return 👍

test/specs/lwcLsp.e2e.ts
```diff
@@ -0,0 +1,124 @@
...
+    }
+    await utilities.pause(5);
+    await utilities.runCommandFromCommandPrompt(
+      workbench,
+      'Developer: Reload Window',
+      10
+    );
+
+    // Verify Lightning Web Components extension is present and running
+    const extensionNameDivs = await $$('div.name');
```
randi274, line 56: Also can you move this out to a utility function? Seems like a useful bit of code, and I see you're using almost the same in the aura lsp

test/specs/lwcLsp.e2e.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  step('Verify Extension is Running', async () => {
+    // Using the Command palette, run Developer: Show Running Extensions
+    const workbench = await browser.getWorkbench();
+    await utilities.runCommandFromCommandPrompt(
+      workbench,
+      'Developer: Show Running Extensions',
+      1
+    );
+    let buttons = await $$('a.monaco-button.monaco-text-button');
```
randi274, line 33: Can you add a comment about why this was necessary? and for the following button click?

test/specs/lwcLsp.e2e.ts
```diff
@@ -0,0 +1,124 @@
...
+    for (const item of buttons) {
+      const text = await item.getText();
+      if (text.includes('Install and Reload')) {
+        await item.click();
```
randi274, line 37: also can early return

test/specs/lwcLsp.e2e.ts
```diff
@@ -0,0 +1,124 @@
...
+    }
+    expect(extensionWasFound).toBe(true);
+  });
+
+  step('Go to Definition (Javascript)', async () => {
+    // Get open text editor
+    const workbench = await browser.getWorkbench();
+    const editorView = workbench.getEditorView();
+    const textEditor = (await editorView.openEditor('lwc1.js')) as TextEditor;
+    await textEditor.moveCursor(3, 40);
```
randi274, line 72: It was really cool to put a breakpoint here and then watch the cursor move in the IDE 🪄

randi274 (changes requested):
This is looking so great! Extra points for having implemented Autocomplete as well. Overall, can you add a ton more logging? This will make it much easier to work with and troubleshoot in a remote environment. For example, in the test setup there are logs generated at the start and end of every function. I think adding logging to the start/end of every step would be good, and maybe even everywhere you've added a comment. 

Screenshot 2023-05-05 at 4 31 05 PM

test/specs/lwcLsp.e2e.ts
```diff
@@ -0,0 +1,124 @@
...
+    // Verify autocompletion options are present
+    const autocompletionOptions = await $$(
+      'textarea.inputarea.monaco-mouse-cursor-text'
+    );
+    expect(await autocompletionOptions[0].getAttribute('aria-haspopup')).toBe(
+      'true'
+    );
+    expect(
+      await autocompletionOptions[0].getAttribute('aria-autocomplete')
+    ).toBe('list');
```
randi274, line 118: Would it also be possible to select any of the options and see that it gets inserted into the document?

test/utilities/lwcUtils.ts
```diff
@@ -0,0 +1,106 @@
...
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { TextEditor } from 'wdio-vscode-service';
+import { runCommandFromCommandPrompt } from './commandPrompt';
+import { log, pause } from './miscellaneous';
+
+export async function enableLWCExtension(): Promise<void> {
+  // salesforce.salesforcedx-vscode-lwc extension is actually not loading automatically
```
randi274, line 13: Excellent clarification here

test/specs/lwcLsp.e2e.ts
randi274: @CristiCanizales did you have any issues moving this one out?

randi274 (approved):
Thanks for adding all the extra logging! This is much easier to follow along with now 😍

PR #12, opened 2023-05-02
Add screenshots on test failure

randi274:
### What does this PR do?
Attaches screenshots to failed e2e testing runs. See the "[screenshots](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/4865736205)" directory that's uploaded for this failed run, which includes 8 tests, one each for all of the currently failing tests. Also save screenshots for local test runs.

Screenshot 2023-06-05 at 3 22 47 PM

@W-13013918@

### Functionality Before
If a test failed, we had no way of viewing what the screen looked like in a virtual environment.

### Functionality After
We see a screenshot for every failed test!


CristiCanizales (changes requested):
Thank you for this Randi! I think this is going to be very helpful.

I verified that a screenshot is indeed added to the screenshots folder every time a test fails! Although, even if a step fails in different parts in different runs, there'll be only one screenshot saved because it's replaced every time.

As discussed on Slack, I think it'd be helpful to keep them all to be able to review every different case:)

test/wdio.conf.ts
```diff
@@ -10,6 +10,8 @@ import type { VSCodeCapabilities } from 'wdio-vscode-service/dist/types';
 import {
   EnvironmentSettings
 } from './environmentSettings';
+import { join } from 'path';
+import { existsSync, mkdir, mkdirSync } from 'fs';
```
jeffb-sfdc, line 14: `mkdir` isn't used and can be removed

jeffb-sfdc:
@randi274 just a heads up, the contents of `wdio.conf.ts` appear to be pretty old.  I suspect this branch needs to pull `develop` into it and get caught up.

test/wdio.conf.ts
```diff
@@ -307,8 +309,18 @@ export const config: Options.Testrunner = {
...
+    if (!passed) {
+      const screenshotDir = join(__dirname, 'screenshots');
+      console.log(`Test run failed. Saving a screenshot of the failure here: ${screenshotDir}`);
+      if (!existsSync(screenshotDir)) {
+        mkdirSync(screenshotDir, { recursive: true });
+      }
+      const sanitizedTestTitle = test.title.replace(/[^a-zA-Z0-9 ]/g, "");
+      await browser.saveScreenshot(join(screenshotDir, `${test.parent} - ${sanitizedTestTitle}.png`))
+    }
+  },
```
jeffb-sfdc, line 323: @randi274 I think this would be cool if it was a utility function, and tests could (also) call it when we need to debug something in GHA.  I'm thinking:
1. create a new file, `/test/utilities/screenshot.ts`
2. within the file, add:
```
export async function saveScreenshot(parentPath: string, testTitle: string): Promise<void> {
  const screenshotDir = join(__dirname, 'screenshots');
  console.log(`Test run failed. Saving a screenshot of the failure here: ${screenshotDir}`);
  if (!existsSync(screenshotDir)) {
    mkdirSync(screenshotDir, { recursive: true });
  }
  const sanitizedTestTitle = testTitle.replace(/[^a-zA-Z0-9 ]/g, "");
  const screenshotPath = join(screenshotDir, `${parentPath} - ${sanitizedTestTitle}.png`);
  await browser.saveScreenshot(screenshotPath);
}
```
(warning-this is untested code)
3. and then in wdio.conf.ts:
```
  // eslint-disable-next-line @typescript-eslint/no-misused-promises
  afterTest: async (test, __, { passed }) => {
    if (passed) {
      return;
    }    

    await utilities.saveScreenshot(test.parent, test.title);
  }
```

randi274: Nice! I think this would fit more with the conventions for the rest of the repo, so I'll try moving this one out as well.

Building on this, I could also distinguish between a screenshot for a test failure and a screenshot we want to save to make debugging easier.

randi274: Also interesting thing I learned - breakpoints work in the utility function called from the afterTest hook, but they *don't* work in the wdio.conf.ts

test/wdio.conf.ts
```diff
@@ -307,8 +309,18 @@ export const config: Options.Testrunner = {
...
-  // },
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises
+  afterTest: async (test, __, { passed }) => {
+    if (!passed) {
+      const screenshotDir = join(__dirname, 'screenshots');
+      console.log(`Test run failed. Saving a screenshot of the failure here: ${screenshotDir}`);
+      if (!existsSync(screenshotDir)) {
+        mkdirSync(screenshotDir, { recursive: true });
+      }
+      const sanitizedTestTitle = test.title.replace(/[^a-zA-Z0-9 ]/g, "");
```
jeffb-sfdc, line 320: For the sanitization, we have tests which have a `-` character in them, so this character should be added.  the RegEx should probably also keep/not strip out `'` and `"` as well.

test/utilities/screenshot.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+/**
+ * Supports multiple runs and indicates when we're receiving the same error message
+ * repeatedly in the same test.
+ * @param saveDir directory where screenshot will be saved
+ * @param error result from the test run; message is an optional param
+ * @returns 
+ */
+function getScreenshotTitle(saveDir: string, error: Error): string {
```
jeffb-sfdc, line 29: Since only the `error.message` property of the `error` object is used, I suggest passing in only what is needed, the message:
```

(line 18)
const screenshotPath = getScreenshotTitle(saveDir, error.message);

function getScreenshotTitle(saveDir: string, errorTitle = 'Unknown error'): string {
  let sanitizedTestTitle = errorTitle.replace(/[^a-zA-Z0-9 ]/g, '');
  const savedScreenshots = readdirSync(saveDir);
.
.
.
```

test/utilities/screenshot.ts
```diff
@@ -0,0 +1,40 @@
...
+  const screenshotPath = getScreenshotTitle(saveDir, error);
+  await browser.saveScreenshot(screenshotPath);
+}
+
+/**
+ * Supports multiple runs and indicates when we're receiving the same error message
+ * repeatedly in the same test.
+ * @param saveDir directory where screenshot will be saved
+ * @param error result from the test run; message is an optional param
+ * @returns
```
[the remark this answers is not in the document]
jeffb-sfdc: But just for the record, I'm not a fan of adding JavaDoc-style comments, so I'm good if they were removed, but if you want them in, that's fine too.

test/utilities/screenshot.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Supports multiple runs and indicates when we're receiving the same error message
+ * repeatedly in the same test.
+ * @param saveDir directory where screenshot will be saved
+ * @param error result from the test run; message is an optional param
+ * @returns 
+ */
+function getScreenshotTitle(saveDir: string, error: Error): string {
+  const errorTitle = error?.message ? error.message : 'Unknown error';
+  let sanitizedTestTitle = errorTitle.replace(/[^a-zA-Z0-9 ]/g, "");
```
jeffb-sfdc, line 31: For the regex, `-` needs to be added, and for good measure, I'd add `'` and `"` too.

test/utilities/screenshot.ts
```diff
@@ -0,0 +1,40 @@
...
+  const savedScreenshots = readdirSync(saveDir);
+
+  const numScreenshotsWithSameError = savedScreenshots?.filter(title => {
+    return title.includes(sanitizedTestTitle);
+  }).length;
+  sanitizedTestTitle = numScreenshotsWithSameError + ' - ' + sanitizedTestTitle;
+
+  return join(saveDir, `${sanitizedTestTitle}.png`);
```
jeffb-sfdc, line 39: This is a minor comment.  Since a string concatenation is already taking place, you could replace lines 37-39 with:
```
  sanitizedTestTitle = numScreenshotsWithSameError + ' - ' + sanitizedTestTitle + '.png';

  return join(saveDir, sanitizedTestTitle);
```

or
```
  sanitizedTestTitle = `${numScreenshotsWithSameError} - ${sanitizedTestTitle}.png`;

  return join(saveDir, sanitizedTestTitle);
```

I'm good with either option1 (the original), option2, or option3.  I just wanted to raise the issue.

test/utilities/screenshot.ts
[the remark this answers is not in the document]
randi274: These ones are hopefully pretty straightforward, so I'll go ahead and remove 👍

test/utilities/screenshot.ts
randi274: ah, yes good catch!

test/utilities/screenshot.ts
randi274: for the regex, I didn't see any dashes in our `step` names, nor quotations. I can't find the doc link to it now, but the reason this was truncated down so harshly was because github actions puts some limitations on what can be saved as part of the uploadArtifact, and I'd rather us be more conservative to get those titles every time than include quotes. (Previously our use of the colon from several `SFDX: ` commands was causing issues!)

test/utilities/screenshot.ts
```diff
@@ -0,0 +1,27 @@
...
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { existsSync, mkdirSync } from 'fs';
+import { join } from 'path';
+import { EnvironmentSettings } from '../environmentSettings';
+
+export async function saveFailedTestScreenshot(specTitle: string, testTitle: string, error: Error): Promise<void> {
+  const saveDir = join(__dirname, '..', '..', 'screenshots', specTitle);
+  console.log(`Test run failed! Failure: ${error?.message}`);
```
jeffb-sfdc, line 14: if error is options, I think it should be:
```
export async function saveFailedTestScreenshot(specTitle: string, testTitle: string, error: Error?): Promise<void> {
  const saveDir = join(__dirname, '..', '..', 'screenshots', specTitle);
  if (error && error.message) {
  console.log(`Test run failed! Failure: ${error?.message}`);
}
```

jeffb-sfdc (approved):

jeffb-sfdc:
Approved

CristiCanizales (approved):
Looks good :) Thank you for adding the time to the screenshots 

🚢

PR #14, opened 2023-05-16
Cristi/visualforce lsp

CristiCanizales:
### What does this PR do?
Implementation for E2E Visualforce LSP

@W-13068545@


test/specs/anInitialSuite.e2e.ts
```diff
@@ -91,59 +103,68 @@ describe('An Initial Suite', async () => {
...
+      await utilities.runCommandFromCommandPrompt(
+        workbench,
+        'Developer: Show Running Extensions',
+        2
+      );
+
+      // Close panel so the visualforce extension can be seen
+      await utilities.runCommandFromCommandPrompt(
+        workbench,
+        'View: Close Panel',
```
CristiCanizales, line 119: @jeffb-sfdc I added this here to close the output panel so we can verify all 7 extensions :)

test/environmentSettings.ts
```diff
@@ -28,32 +28,47 @@ export class EnvironmentSettings {
     // './test/specs/**/sObjectsDefinitions.e2e.ts'
     // './test/specs/**/templates.e2e.ts',
     // './test/specs/**/trailApexReplayDebugger.e2e.ts',
-    // './test/specs/**/visualForceLsp.e2e.ts',
+    // './test/specs/**/visualforceLsp.e2e.ts'
```
jeffb-sfdc, line 31: Good catch!  I thought it was "VisualForce".

test/specs/anInitialSuite.e2e.ts
jeffb-sfdc: Nice!  All are displayed on my monitor, but several other extension also appear, so this might flap in GitHub, but for the moment I say we keep what you have.

test/specs/runApexTests.e2e.ts
```diff
@@ -35,12 +33,23 @@ describe('Run Apex Tests', async () => {
     // Create Apex class 3 and test
     await utilities.createApexClassWithTest('ExampleApexClass3');
 
-    // Push source to scratch org
+    // Push source to org
```
jeffb-sfdc, line 36: This is a scratch org, so I think the comment on line 38/36 should remain with "scratch" in it.

test/specs/runApexTests.e2e.ts
```diff
@@ -146,32 +225,47 @@ describe('Run Apex Tests', async () => {
...
+    )) as TextEditor;
     const testText = [
       '@IsTest',
       'public class ExampleApexClass1Test {',
       '\t@IsTest',
       '\tstatic void validateSayHello() {',
-      '\t\tSystem.debug(\'Starting validate\');',
-      '\t\tExampleApexClass1.SayHello(\'Andres\');',
-      '\t\tSystem.assertEquals(1, 1, \'all good\');',
+      "\t\tSystem.debug('Starting validate');",
```
jeffb-sfdc, line 236: I like the other way, with single quotes, and then escaping the inner quotes better:
```
const testText = [
      '@IsTest',
      'public class ExampleApexClass1Test {',
      '\t@IsTest',
      '\tstatic void validateSayHello() {',
      '\t\tSystem.debug(\'Starting validate\');',
      '\t\tExampleApexClass1.SayHello(\'Andres\');',
      '\t\tSystem.assertEquals(1, 1, \'all good\');',
      '\t}',
      '}'
    ].join('\n');
```

test/specs/runApexTests.e2e.ts
```diff
@@ -310,74 +484,154 @@ describe('Run Apex Tests', async () => {
     // Create Apex class AccountService
     await utilities.createApexClassWithBugs();
 
-    // Push source to scratch org
+    // Push source to org
```
jeffb-sfdc, line 487: Please add "scratch" back in

test/specs/sObjectsDefinitions.e2e.ts
```diff
@@ -6,9 +6,7 @@
  */
 import { step } from 'mocha-steps';
 import { DefaultTreeItem, TreeItem } from 'wdio-vscode-service';
-import {
-  TestSetup
-} from '../testSetup';
+import { TestSetup } from '../testSetup';
```
jeffb-sfdc, line 9: The import statements are out of order.
```
import * as fs from 'fs-extra';
import { step } from 'mocha-steps';
import path from 'path';
import { DefaultTreeItem, TreeItem } from 'wdio-vscode-service';
import { TestSetup } from '../testSetup';
import * as utilities from '../utilities';
```

test/specs/sObjectsDefinitions.e2e.ts
```diff
@@ -24,10 +22,22 @@ describe('SObjects Definitions', async () => {
...
+    );
+    const destination = path.join(
+      projectPath!,
+      'force-app',
+      'main',
+      'default',
+      'objects'
+    );
+
+    fs.copy(source, destination, { recursive: true }, async error => {
```
jeffb-sfdc, line 40: this PR will need to go in before your PR for Manifest Builder

test/specs/visualforceLsp.e2e.ts
```diff
@@ -0,0 +1,136 @@
...
+import { TestSetup } from '../testSetup';
+import * as utilities from '../utilities';
+
+describe('Visualforce LSP', async () => {
+  let testSetup: TestSetup;
+  let projectName: string;
+  const fiveMinutes = 5 * 60;
+
+  step('Set up the testing environment', async () => {
+    utilities.log(`VisualforceLsp - Set up the testing environment`);
```
jeffb-sfdc, line 18: Since there are no string substitutions, the backtick isn't needed here, and just a "single quote string' should be used.

test/utilities/apexUtils.ts
```diff
@@ -178,3 +177,78 @@ export async function createAnonymousApexFile(): Promise<void> {
   await textEditor.save();
   await pause(1);
 }
+
+export async function createVisualforcePage(): Promise<void> {
```
jeffb-sfdc, line 181: `createVisualforcePage()` is in apexUtils.ts.  It seems like this isn't Apex and instead should be in a new file, visualforceUtils.ts.

test/specs/visualforceLsp.e2e.ts
```diff
@@ -0,0 +1,136 @@
...
+      'FooPage.page'
+    )) as TextEditor;
+    await textEditor.typeTextAt(3, 1, '\t<apex:pageM');
+    await utilities.pause(1);
+
+    // Verify autocompletion options are present
+    const autocompletionOptions = await $$(
+      'textarea.inputarea.monaco-mouse-cursor-text'
+    );
+    expect(await autocompletionOptions[0].getAttribute('aria-haspopup')).toBe(
```
jeffb-sfdc, line 114: I'm not a fan of this style, since it makes debugging more difficult.  I suggest splitting it up and changing to:
```
    const ariaHasPopupAttribute = await autocompletionOptions[0]
      .getAttribute('aria-haspopup');
    expect(ariaHasPopupAttribute).toBe('true');

    const ariaAutocompleteAttribute = await autocompletionOptions[0]
      .getAttribute('aria-autocomplete');
    expect(ariaAutocompleteAttribute).toBe('list');
```

test/specs/visualforceLsp.e2e.ts
```diff
@@ -0,0 +1,136 @@
...
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { step } from 'mocha-steps';
+import { TextEditor } from 'wdio-vscode-service';
+import { TestSetup } from '../testSetup';
+import * as utilities from '../utilities';
+
+describe('Visualforce LSP', async () => {
+  let testSetup: TestSetup;
+  let projectName: string;
```
jeffb-sfdc, line 14: `projectName` isn't used and can be removed

test/utilities/apexUtils.ts
```diff
@@ -178,3 +177,78 @@ export async function createAnonymousApexFile(): Promise<void> {
...
+  // Select the default directory (press Enter/Return).
+  await inputBox.confirm();
+  await pause(1);
+
+  // Modify page content
+  textEditor = (await editorView.openEditor('FooPage.page')) as TextEditor;
+  const pageText = [
+    `<apex:page controller="myController" tabStyle="Account">`,
+    `\t<apex:form>`,
+    `\t`,
```
jeffb-sfdc, line 243: Nit pick: line 243 inserts an extra blank line, and should probably be removed.

test/specs/visualforceLsp.e2e.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  step('Autocompletion', async () => {
+    utilities.log(`${testSetup.testSuiteSuffixName} - Autocompletion`);
+    // Get open text editor
+    const workbench = await browser.getWorkbench();
+    const editorView = workbench.getEditorView();
+    const textEditor = (await editorView.openEditor(
+      'FooPage.page'
+    )) as TextEditor;
+    await textEditor.typeTextAt(3, 1, '\t<apex:pageM');
```
jeffb-sfdc, line 107: nit pick: looks like an additional tab should be added.  Yea, I know know it's not needed.. ;-)

test/specs/runApexTests.e2e.ts
CristiCanizales: Uh good catch, I can't remember having changed this on purpose, I think it was the formatter. I'll change this to be the same way as in apexUtils

test/utilities/apexUtils.ts
CristiCanizales: I think I rather like it here. We won't be doing (much) else visualforce related and we're actually creating the apex controller for this as well, so I don't think an extra file is necessary and I'd leave it here

test/specs/runApexTests.e2e.ts
CristiCanizales: I'd leave it without "scratch" for consistency, and it's no wrong anyways, in the end it's an org
image

jeffb-sfdc: OK

test/utilities/apexUtils.ts
```diff
@@ -178,3 +177,78 @@ export async function createAnonymousApexFile(): Promise<void> {
   await textEditor.save();
   await pause(1);
 }
+
+export async function createVisualforcePage(): Promise<void> {
+  const workbench = await browser.getWorkbench();
+  let textEditor: TextEditor;
```
jeffb-sfdc, line 183: `textEditor` isn't needed until line 202.  Could you remove line 183, and then add 'let' to line 202?

test/utilities/apexUtils.ts
jeffb-sfdc: They aren't the same entity type, and should be in different files.  What I'd like to see is:
1. lines 185-221 refactored into it's own function.  Maybe call it, "createApexConrtoller()"
2. create visualforceutils.ts and move createVisualforcePage() into this file
3. add expectations/validations for what createVisualforcePage() created.  Eg - validate that /pages/FooPage.page and /pages/FooPage.page-meta.xml were created, and inside FooPage.page's editor, validate that `<apex:page>` and `</apex:page>` exist

test/specs/visualforceLsp.e2e.ts
```diff
@@ -0,0 +1,137 @@
...
+    expect(line3Text).toContain('apex:pageMessage');
+  });
+
+  step('Tear down and clean up the testing environment', async () => {
+    utilities.log(
+      `${testSetup.testSuiteSuffixName} - Tear down and clean up the testing environment`
+    );
+    await testSetup.tearDown();
+  });
+});
```
jeffb-sfdc, line 137: It looks like the "SFDX: Create Visualforce Page" command is covered, but the "SFDX: Create Visualforce Component" command isa lacking coverage.

CristiCanizales: Yup, that should be covered in the templates suite 
image

test/utilities/apexUtils.ts
CristiCanizales: I think point 3 is out of scope for this suite/PR and is something to be validated in templates.e2e.ts, to validate the create command. Here we just have the utility, validation should be done in that other file.

test/specs/visualforceLsp.e2e.ts
jeffb-sfdc: Ah, got it, thanks.  At the time (this morning) I was thinking it was to be here in this suite.

test/utilities/apexUtils.ts
jeffb-sfdc: If no checks/expectations within `createVisualforcePage()`, then in the step() in visualforceLsp.e2e, right after `createVisualforcePage()` is called.

Remember, when this runs in GHA, it's going to be a black box and will be difficult (and expensive) to debug when something goes wrong.  What happens if createVisualforcePage() fails - when it doesn't work, but no exceptions either?  It will fail later on, but will be difficult to know where it really as failing.

PR #15, opened 2023-05-16
Cristi/apex lsp

CristiCanizales:
### What does this PR do?
Implementation for E2E Apex LSP

@W-13068559@

### DIFF
--- a/test/specs/apexLsp.e2e.ts
+++ b/test/specs/apexLsp.e2e.ts
@@ -15,7 +15,7 @@ describe('Apex LSP', async () => {
 
   step('Set up the testing environment', async () => {
     utilities.log(`ApexLsp - Set up the testing environment`);
-    testSetup = new TestSetup('ApexLsp', true);
+    testSetup = new TestSetup('ApexLsp', false);
     await testSetup.setUp();
     projectName = testSetup.tempProjectName.toUpperCase();
 
--- a/test/specs/apexLsp.e2e.ts
+++ b/test/specs/apexLsp.e2e.ts
@@ -1,0 +1,110 @@
+/*
+ * Copyright (c) 2023, salesforce.com, inc.
+ * All rights reserved.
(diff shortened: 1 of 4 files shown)


test/specs/apexLsp.e2e.ts
```diff
@@ -5,29 +5,106 @@
...
-    // TODO: implement
-    expect(1).toBe(1);
+  step('Go to Definition', async () => {
+    utilities.log(`${testSetup.testSuiteSuffixName} - Go to Definition`);
+    // Get open text editor
+    const workbench = await browser.getWorkbench();
+    const editorView = workbench.getEditorView();
+    const textEditor = (await editorView.openEditor(
+      'ExampleClassTest.cls'
+    )) as TextEditor;
```
lahernandezb, line 57: Hey Cristi, this looks good. More of a minor suggestion and less of a requirement would be to place this setup piece in a util or a pre hook to avoid repetition. I suspect this will be used quite frequently.

test/testSetup.ts
```diff
@@ -15,16 +15,13 @@ import {
   QuickOpenBox,
   Workbench
 } from 'wdio-vscode-service';
-import {
-  EnvironmentSettings
-} from './environmentSettings';
+import { EnvironmentSettings } from './environmentSettings';
```
lahernandezb, line 18: All the changes in this file seem to be spacing changes. Is your editor using a abiding by the `.prettierrc`?

lahernandezb (approved):
Cristi this looks good so I approved, but left a minor suggestion if you feel up for it.

test/specs/apexLsp.e2e.ts
CristiCanizales: I think we do use it frequently but also sometimes for different purposes, and wdio is just weird so I'd keep it in how it is in every step needed at least for now

PR #21, opened 2023-06-07
fix: add ability to use an existing sfdx project

peternhale:
@W-13190794@

Add ability to use an existing sfdx prject for tests

Adding this to support perf testing of product.

### DIFF
--- a/test/specs/apexLspPerf.perf.ts
+++ b/test/specs/apexLspPerf.perf.ts
@@ -24,6 +24,8 @@ describe('Apex LSP Perf', async () => {
 
     await utilities.runCommandFromCommandPrompt(workbench, 'Developer: Show Running Extensions', 1);
 
+    await utilities.runCommandFromCommandPrompt(workbench, 'SFDX: Create Project', 10);
+
     // Verify Apex extension is present and running
     const extensionWasFound = await utilities.findExtensionInRunningExtensionsList(
       workbench,
--- a/test/testSetup.ts
+++ b/test/testSetup.ts
@@ -17,7 +17,7 @@ export class TestSetup {
   public testSuiteSuffixName: string;
   private reuseScratchOrg = false;
   private static aliasAndUserNameWereVerified = false;
-  public tempFolderPath: string | undefined = undefined;
+  public tempFolderPath: string;
   protected _projectFolderPath: string | undefined = undefined;
   private prompt: QuickOpenBox | InputBox | undefined;
   private scratchOrgAliasName: string | undefined;
@@ -28,6 +28,7 @@ export class TestSetup {
   ) {
     this.testSuiteSuffixName = testSuiteSuffixName;
     this.reuseScratchOrg = reuseScratchOrg;
+    this.tempFolderPath = path.join(__dirname, '..', 'e2e-temp');
   }
 
   public get tempProjectName(): string {
@@ -65,7 +66,6 @@ export class TestSetup {
     utilities.log('');
     utilities.log(`${this.testSuiteSuffixName} - Starting setUpTestingEnvironment()...`);
 
-    this.tempFolderPath = path.join(__dirname, '..', 'e2e-temp');
     this.projectFolderPath = path.join(this.tempFolderPath, this.tempProjectName);
 
     utilities.log(`${this.testSuiteSuffixName} - creating project files in ${this.projectFolderPath}`);
@@ -74,13 +74,17 @@ export class TestSetup {
     await this.removeProject();
 
     // Now create the temp folder.  It should exist but create the folder if it is missing.
+    await this.createTmpFolder();
+
+    utilities.log(`${this.testSuiteSuffixName} - ...finished setUpTestingEnvironment()`);
+    utilities.log('');
+  }
+
+  protected async createTmpFolder() {
     if (!fs.existsSync(this.tempFolderPath)) {
       await utilities.createFolder(this.tempFolderPath);
       await utilities.pause(1);
     }
-
-    utilities.log(`${this.testSuiteSuffixName} - ...finished setUpTestingEnvironment()`);
-    utilities.log('');
   }
 
   public async createProject(): Promise<void> {
--- a/test/testSetupWithExistingProject.ts
+++ b/test/testSetupWithExistingProject.ts
@@ -16,4 +16,9 @@ export class TestSetupWithWExistingProject extends TestSetup {
   protected removeProject(): Promise<void> {
     return Promise.resolve();
   }
+
+  // overriding base class method to not create a tmp folder
+  protected createTmpFolder(): Promise<void> {
+    return Promise.resolve();
+  }
 }


test/environmentSettings.ts
```diff
@@ -6,16 +6,17 @@
...
 export class EnvironmentSettings {
   private static _instance: EnvironmentSettings;
 
   private _vscodeVersion = 'stable'; //  or 'insiders' or '1.77.3'
+  private _sfdxProjectHome = join(__dirname, '..', '..');
   private _specFiles = [
     './test/specs/**/*.e2e.ts'
     // OR
     // './test/specs/**/anInitialSuite.e2e.ts',
-    // './test/specs/**/apexLsp.e2e.ts',
```
CristiCanizales, line 18: Why is this deleted?

CristiCanizales (commented):
LGTM :) I just have a question... Why is the content of steps 1 and 2 commented?

test/testSetup.ts
```diff
@@ -82,16 +95,18 @@ export class TestSetup {
     );
 
     this.tempFolderPath = path.join(__dirname, '..', 'e2e-temp');
-    this.projectFolderPath = path.join(
-      this.tempFolderPath,
-      this.tempProjectName
-    );
+    // set projectFolderPath if it is undefined
+    this.projectFolderPath ??=
```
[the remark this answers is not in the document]
peternhale: The original syntax is correct `??=` per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment

test/environmentSettings.ts
peternhale: It should not be.

peternhale:
> LGTM :) I just have a question... Why is the content of steps 1 and 2 commented?

The commented code will be next steps. I need to get through the first pass on being able to run the test successfully with yourkit profiling enabled. I can remove it if that makes better sense.

CristiCanizales (approved):
🚢

test/environmentSettings.ts
```diff
@@ -6,11 +6,13 @@
  */
 
 import { join } from 'path';
+import * as utilities from './utilities';
 
 export class EnvironmentSettings {
   private static _instance: EnvironmentSettings;
 
   private _vscodeVersion = 'stable'; //  or 'insiders' or '1.77.3'
+  private _sfdxProjectHome = join(__dirname, '..', '..');
```
jeffb-sfdc, line 15: I think `join(__dirname, '..', '..', 'apex-perf-project');` would be more appropriate for `_sfdxProjectHome`.

Also, I think `_sfdxTestProjectLocation` or `_apexPerfProjectLocation` would be a more appropriate name.

test/environmentSettings.ts
```diff
@@ -55,10 +57,13 @@ export class EnvironmentSettings {
...
 
-      if (process.env.SPEC_FILES) {
-        EnvironmentSettings._instance._specFiles = [
-          './test/specs/**/' + process.env.SPEC_FILES
-        ];
+      if (process.env.SPEC_FILES?.length) {
+        EnvironmentSettings._instance._specFiles = process.env.SPEC_FILES.split(
+          ','
+        ).map(specFile => `./test/specs/**/${specFile.trim()}`);
+        utilities.log(
```
jeffb-sfdc, line 64: `getInstance()` is starting to get a bit long.  I'm thinking maybe the initialization code should be moved into the constructor:

```

  private constructor() {
    this._vscodeVersion =
      process.env.VSCODE_VERSION ||
      EnvironmentSettings._instance._vscodeVersion;

    if (process.env.SPEC_FILES?.length) {
      this._specFiles = process.env.SPEC_FILES.split(',').map(
        specFile => `./test/specs/**/${specFile.trim()}`
      );
      utilities.log(`SPEC_FILES: ${this._specFiles.join(', ')}`);
    }

    this._devHubAliasName =
      process.env.DEV_HUB_ALIAS_NAME || this._devHubAliasName;

    this._devHubUserName =
      process.env.DEV_HUB_USER_NAME || this._devHubUserName;

    this._extensionPath = process.env.EXTENSION_PATH || this._extensionPath;

    this._throttleFactor =
      parseInt(process.env.THROTTLE_FACTOR!) || this._throttleFactor;

    if (process.env.WITH_TEST_PROJECT) {
      this._sfdxProjectHome = join(
        __dirname,
        '..',
        '..',
        process.env.WITH_TEST_PROJECT
      );
    }
  }

  public static getInstance(): EnvironmentSettings {
    if (!EnvironmentSettings._instance) {
      EnvironmentSettings._instance = new EnvironmentSettings();
    }

    return EnvironmentSettings._instance;
  }
```

thoughts?

test/environmentSettings.ts
```diff
@@ -73,6 +78,14 @@ export class EnvironmentSettings {
       EnvironmentSettings._instance._throttleFactor =
         parseInt(process.env.THROTTLE_FACTOR!) ||
         EnvironmentSettings._instance._throttleFactor;
+      if (process.env.WITH_TEST_PROJECT) {
+        EnvironmentSettings._instance._sfdxProjectHome = join(
+          __dirname,
+          '..',
+          '..',
+          process.env.WITH_TEST_PROJECT
```
jeffb-sfdc, line 86: Since this is just the name of the test project (apex-perf-project), `TEST_PROJECT_NAME` or `SFDX_TEST_PROJECT_NAME` seems more clear.

test/testSetup.ts
```diff
@@ -109,6 +124,11 @@ export class TestSetup {
   }
 
   public async createProject(scratchOrgEdition: string): Promise<void> {
+    if (this.usingExistingProject) {
+      utilities.log('Using existing project, skipping createProject()');
+      return;
+    }
+
```
jeffb-sfdc, line 131: Having read through the PR, I think TestSetup is too muddy, and trying to do something it wasn't meant to do.

I think what would be better is:
1. revert the changes in testSetup.ts
2. in apexLspPerf.ts, in the first step, change it to:
```
  step('Ensure Apex Language Server Starts Successfully', async () => {
    utilities.log('ApexLsp - Set up the testing environment');
    testSetup = new TestSetup(
      'ApexLsp',
      false
    );

    await testSetup.setUpTestingEnvironment();
    .
    .
    .
```

In this suite, after `testSetup.setUpTestingEnvironment()` is called, the other functions in `setUp()` (`createProject()`, `authorizeDevHub()`, and `createDefaultScratchOrg()`) aren't used, so I think apexLspPerf.ts should manage its setup and tear down, and not `TestSetup`.

test/specs/apexLspPerf.perf.ts
```diff
@@ -0,0 +1,112 @@
...
+    const workbench = await browser.getWorkbench();
+
+    await utilities.runCommandFromCommandPrompt(
+      workbench,
+      'Developer: Show Running Extensions',
+      1
+    );
+
+    // Verify Apex Language Server extension is present and running
+    const extensionNameDivs = await $$('div.name');
```
jeffb-sfdc, line 35: Lines 35-42: The latest version of the `develop` branch now has `/utilities/extensionUtils.ts` and has `findExtensionInRunningExtensionsList()`

test/utilities/fileSystem.ts
```diff
@@ -5,22 +5,17 @@
...
 import path from 'path';
 import { TestSetup } from '../testSetup';
 import { log } from './miscellaneous';
 
-export function createFolder(folderPath: string): ChildProcess {
-  const childProcess = exec(`mkdir "${folderPath}"`);
-
-  return childProcess;
+export function createFolder(folderPath: string): Promise<void> {
+  return fs.mkdirp(folderPath);
```
jeffb-sfdc, line 14: Good find!  The APIs in the `fs` module are probably faster than `exec`.  I'm trying to recall if there was a reason why I used `exec` and `ChildProcess` but I can't think of any.  Well, if running all of the suites pass, then I say let's switch.

package.json
```diff
@@ -18,7 +18,7 @@
     "mocha-steps": "^1.3.0",
     "prettier": "1.19.1",
     "ts-node": "^10.9.1",
-    "typescript": "^4.9.4",
+    "typescript": "^4.9.5",
```
jeffb-sfdc, line 21: The `package-lock.json` file also needs to be checked in.

test/environmentSettings.ts
peternhale: I am consider these changes not for the purpose of satisfying my needs, but for the needs that follow.
Just trying to satisfy the need for an alternative project.

test/environmentSettings.ts
peternhale: I'll move it up into ctor

test/environmentSettings.ts
peternhale: I like TEST_PROJECT_NAME, will change

package.json
peternhale: Will do

test/utilities/fileSystem.ts
peternhale: This gets us closer to platform independence (thinking Windows here)

test/testSetup.ts
```diff
@@ -109,6 +124,11 @@ export class TestSetup {
   }
 
   public async createProject(scratchOrgEdition: string): Promise<void> {
+    if (this.usingExistingProject) {
```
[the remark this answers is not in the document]
peternhale: testSetup is predicated on a single use case, our current e2e test needs. I see the entire repo as a possibility for supporting not just e2e for vsce, but others that need to test extensions of there own that are to use in conjunction with dx projects.

The parallel I am drawing is to [CLI Plugins Testkit](https://github.com/salesforcecli/cli-plugins-testkit/blob/HEAD/SAMPLES.md), which provides much of the same functionality, but focused on plugin development.

I see this repo as the same offering to extension authors.

I agree that the code I have added here is orthogonal to the meaning of the function name, but I see similar issues with dev hub auth. No check exists to confirm that the env var for hub identity are set.

I will move the guard into setup method, bypassing the call to createProject.

test/testSetup.ts
peternhale: IMO, the long game for me for TestSetup that will serve the test writer w/o having to decide if it can even be used or not.

I know this module is still under heavy development and the e2e test success is paramount, but I really would prefer not to do my own setup/teardown of what seems to me as a very common use case.

test/testSetup.ts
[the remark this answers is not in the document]
jeffb-sfdc: _"...I see the entire repo as a possibility for supporting not just e2e for vsce, but others that need to test extensions of there own that are to use in conjunction with dx projects"_

The repo _is specifically_ for e2e testing and was not designed to be a generic testing platform.  If we can add (with little overhead and minor changes) support for perf testing, I'm all for that.  However, I do not want to design it for something where a specific requirement and a specific goal do not exist.

At the moment, in apexLspPerf.ts...
`testSetup.setUpTestingEnvironment()` is called
and `testSetup.tearDown()` is also called...
...however, `tearDown()` runs "sfdx force:org:delete" on the scratch org that was created, but in this suite, a scratch org is not created, so `tearDown()` should not be called.

...so
1. The majority of TestSetup is not used by apexLspPerf.ts
2. The only function that's used is `setUpTestingEnvironment()`
3. The majority of the proposed changes do not seem appropriate for this class
4. We're trying to make TestSetup do something it was not meant or designed to do
5. The proposed changes break (IMHO) the single responsibility of the class.

I've been thinking about this, and here's what I suggest:
1. back out the changes to testSetup.ts
2. then, in testSetup.ts, add:
```
  public projectExists(): boolean {
    return fs.existsSync(this.projectFolderPath);
  }
```
...and in `public async setUpTestingEnvironment()`, at line 94 of the existing version, change
```
if (fs.existsSync(this.projectFolderPath)) {
```
...to:
```
if (this.projectExists()) {
```

4. create a new file, perfSetup.ts and have it extend testSetup.ts, implement its own setUpTestingEnvironment() and its own projectExists():
```

export class PerfSetup extends TestSetup {
  public projectFolderPath: string | undefined = undefined;
  public usingExistingProject: boolean;

  public constructor(
    testSuiteSuffixName: string,
    reuseScratchOrg: boolean,
    existingProjectPath?: string
  ) {
    super(testSuiteSuffixName, reuseScratchOrg);

    if (existingProjectPath && !fs.existsSync(existingProjectPath)) {
      throw new Error(
        `In TestSetup(), the existing project path '${existingProjectPath}' does not exist`
      );
    }

    this.usingExistingProject = existingProjectPath !== undefined;
    this.projectFolderPath = existingProjectPath;
  }

  public async setUpTestingEnvironment(): Promise<void> {
    // set projectFolderPath if it is undefined
    this.projectFolderPath ??=
      path.join(this.tempFolderPath, this.tempProjectName);

    utilities.log(
      this.usingExistingProject
        ? `${this.testSuiteSuffixName} - using existing project at ${this.projectFolderPath}`
        : `${this.testSuiteSuffixName} - creating project files in ${this.projectFolderPath}`
    );

    super.setUpTestingEnvironment();
  }

  public projectExists(): boolean {
    return fs.existsSync(this.projectFolderPath) && !this.usingExistingProject;
  }
}
```

This way, perfSetup can use the parts of testSetup it wants, it can customize what it wants, and testSetup remains mostly unaffected.


Or, alternatively, we could create a new project and repo specifically for perf testing.  Creating a new wdio-vscode-service project is pretty easy - just run `npm create wdio ./`

See https://www.npmjs.com/package/wdio-vscode-service for more info on this.

peternhale:
> _"...I see the entire repo as a possibility for supporting not just e2e for vsce, but others that need to test extensions of there own that are to use in conjunction with dx projects"_

The repo _is specifically_ for e2e testing and was not designed to be a generic testing platform.  If we can add (with little overhead and minor changes) support for perf testing, I'm all for that.  However, I do not want to design it for something where a specific requirement and a specific goal does not exist.

At the moment, in apexLspPerf.ts...
`testSetup.setUpTestingEnvironment()` is called
and `testSetup.tearDown()` is also called...
...however, `tearDown()` runs "sfdx force:org:delete" on the scratch org that was created, but in this suite, a scratch org is not created, so `tearDown()` should not be called.

...so
1. The majority of TestSetup is not used by apexLspPerf.ts
2. The only function that's used is `setUpTestingEnvironment()`
3. The majority of the proposed changes do not seem appropriate for this class
4. We're trying to make TestSetup do something it was not meant or designed to do
5. The proposed changes break (IMHO) the single responsibility of the class.

I've been thinking about this, and here's what I suggest:
1. back out the changes to testSetup.ts
2. then, in testSetup.ts, add:
```
  public projectExists(): boolean {
    return fs.existsSync(this.projectFolderPath);
  }
```
...and in `public async setUpTestingEnvironment()`, at line 94 of the existing version, change
```
if (fs.existsSync(this.projectFolderPath)) {
```
...to:
```
if (this.projectExists()) {
```

4. create a new file, perfSetup.ts and have it extend testSetup.ts, implement its own setUpTestingEnvironment() and its own projectExists():
```

export class PerfSetup extends TestSetup {
  public projectFolderPath: string | undefined = undefined;
  public usingExistingProject: boolean;

  public constructor(
    testSuiteSuffixName: string,
    reuseScratchOrg: boolean,
    existingProjectPath?: string
  ) {
    super(testSuiteSuffixName, reuseScratchOrg);

    if (existingProjectPath && !fs.existsSync(existingProjectPath)) {
      throw new Error(
        `In TestSetup(), the existing project path '${existingProjectPath}' does not exist`
      );
    }

    this.usingExistingProject = existingProjectPath !== undefined;
    this.projectFolderPath = existingProjectPath;
  }

  public async setUpTestingEnvironment(): Promise<void> {
    // set projectFolderPath if it is undefined
    this.projectFolderPath ??=
      path.join(this.tempFolderPath, this.tempProjectName);

    utilities.log(
      this.usingExistingProject
        ? `${this.testSuiteSuffixName} - using existing project at ${this.projectFolderPath}`
        : `${this.testSuiteSuffixName} - creating project files in ${this.projectFolderPath}`
    );

    super.setUpTestingEnvironment();
  }

  public projectExists(): boolean {
    return fs.existsSync(this.projectFolderPath) && !this.usingExistingProject;
  }
}
```

This way, perfSetup can use the parts of testSetup it wants, it can customize what it wants, and testSetup remains mostly unaffected.


Or, alternatively, we could create a new project and repo specifically for perf testing.  Creating a new wdio-vscode-service project is pretty easy - just run `npm create wdio ./`

See https://www.npmjs.com/package/wdio-vscode-service for more info on this.

test/testSetup.ts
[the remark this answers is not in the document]
peternhale: I am not sure what the contention is. TestSetup already accounts for runtime defined dx projects; via createProject (or, now, an existing one).

The goal these changes, for my current work, is to support existing projects, not specifically perf testing. With this one change others can now use the framework to use other projects, which may come with other needs.

For perf testing I don't need to create a project, create scratch orgs or auth dev hubs.

I am trying to make TestSetup a bit more flexible w/o breaking how its being used now, by adding some guardrails in places folks might run into issues.

I don't see the need to extend TestSetup move the guardrails out of TestSetup. Composition over inheritance is the pattern I prefer and the subclass is substantially different to, IMO, warrant a subclass.

test/environmentSettings.ts
```diff
@@ -33,48 +35,48 @@ export class EnvironmentSettings {
...
+  private _devHubSfdxAuthUrl?: string = undefined;
 
-  private constructor() {}
+  private constructor() {
+    this._vscodeVersion =
+      process.env.VSCODE_VERSION || this._vscodeVersion;
+
+    if (process.env.SPEC_FILES?.length) {
+      this._specFiles = process.env.SPEC_FILES.split(',').map(
+        specFile => `./test/specs/**/${specFile.trim()}`
```
jeffb-sfdc, line 51: @peternhale Just a heads up, with some of the work that @RitamAgrawal is doing, we might be changing the format of "SPEC_FILES"

As it is now, `process.env.SPEC_FILES` holds just the file name, assumes that the files are all in `./test/specs`, and then concatenates the strings.

Ritam is working on something where she might need a suite to be outside this directory location, so we might be changing the code from:
`'./test/specs/**/' + process.env.SPEC_FILES`
(and SPEC_FILES is foo.e2.ts)
...to:
`process.env.SPEC_FILES`
(and SPEC_FILES is whatever the env var is set to, like  ./test/specs/**/foo.e2.ts)

.vscode/launch.json
```diff
@@ -35,6 +35,27 @@
       "console": "integratedTerminal",
       "internalConsoleOptions": "neverOpen"
     },
+    {
+      "name": "(PNH) Debug Current Automation Test",
```
jeffb-sfdc, line 39: We normally don't check in launch configs or tasks which are user specific.  I would suggest changing the launch config a little and make it user-agnostic...
```
 {
      "name": "Debug Current Performance Test",
      "type": "node",
      "request": "launch",
      "program": "${workspaceRoot}/node_modules/@wdio/cli/bin/wdio.js",
      "args": [
        "run",
        "test/wdio.conf.ts",
        "--spec",
        "${file}"
      ],
      "cwd": "${workspaceFolder}",
      "autoAttachChildProcesses": true,
      "console": "integratedTerminal",
      "internalConsoleOptions": "neverOpen"
    },
```

...and then also set the necessary environment variables:
```
(in a terminal window)
$ export EXTENSION_PATH="/Users/peter.hale/git/vse/packages"
$ export DEV_HUB_SFDX_AUTH_URL="force://PlatformCLI::5Aep8612E5JLxJp3ESIXoO277mV4I.qJxA4bxutk8bsnO5B93z3sW1ibxM3WkNhN.lnmlQw6qDjUkIUjUtZ4BC1@na40-dev-hub.my.salesforce.com"
$ export DEV_HUB_USER_NAME="admin@integrationtesthubna40.org"      
```

And if it were me, I'd create a shell script with those three exports in them, and run the script anytime I needed the env vars set.

package.json
```diff
@@ -6,6 +6,7 @@
   "private": true,
   "devDependencies": {
     "@commitlint/config-conventional": "^17.1.0",
+    "@salesforce/core": "^3.36.2",
```
jeffb-sfdc, line 9: I was discussing this PR with @RitamAgrawal and she brought up that one of the original goals and design requirements of the E2E tests, was to:
1. only interact with VS Code like a user would
2. not call any of the internal libraries we use

I see arguments for and against using core here...
For:
The places where salesforce/core are used is when we are automating org auth calls to sfdx, so this isn't UI or user actions that we're automating

Against:
By adding core, this opens the door for APIs to be called instead of controlling VS Code like a user would.

I only am here for a few more days, so I think you & @CristiCanizales  and @CristiCanizales  should discus this.

PR #23, opened 2023-06-15
feat: Ra/add persistent org

RitamAgrawal:
**What does this PR do?**

- Adds a script `setup` which runs `anInitialSetUp.e2e.ts` before running any e2e test
- Adds an e2e test at `test/setup/ anInitialSetUp.e2e.ts` 
- The setup e2e test authorizes to a persistent testing org in remote environment like Github Actions. It authorizes through sfdxAuthUrl.


test/setup/anInitialSetUp.e2e.ts
```diff
@@ -0,0 +1,56 @@
...
+  });
+
+  step('Authorize to Testing Org', async () => {
+    const sfdxAuthUrl = String(SFDX_AUTH_URL);
+    const authFilePath = 'authFile.txt';
+
+    // create and write in a text file
+    fs.writeFileSync(authFilePath, sfdxAuthUrl);
+
+    // For sfdx -> sf, remove above two lines and keep this
```
CristiCanizales, line 39:
```suggestion
    // For sfdx -> sf, remove the two lines below this comment block and uncomment the following line instead
```

test/setup/anInitialSetUp.e2e.ts
```diff
@@ -0,0 +1,56 @@
...
+    // await exec(`sf org login sfdx-url --sfdx-url-file ${authFilePath} --set-default --alias ${devHubAliasName}`);
+    await exec(`sfdx auth:sfdxurl:store -f ${authFilePath}`);
+    await exec(`sfdx alias set ${devHubAliasName}=${devHubUserName}`);
+  });
+
+  step('Verify Connection to the Testing Org', async () => {
+    const workbench = await browser.getWorkbench();
+    const terminalView = await utilities.executeCommand(workbench, `sfdx org list`)
```
CristiCanizales, line 47:
```suggestion
    const terminalView = await utilities.executeCommand(workbench, 'sfdx org list')
```
I recently learnt from @jeffb-sfdc that 'Backtick-style strings perform more processing', so as we don't have to make any string substitutions here, it'd be more appropriate to use `'` (or `"`) here 😄

test/setup/anInitialSetUp.e2e.ts
```diff
@@ -0,0 +1,56 @@
...
+    const terminalText = await utilities.getTerminalViewText(terminalView, 100);
+
+    expect(terminalText).toContain(`${orgId}`);
+    expect(terminalText).toContain('Connected');
+    expect(terminalText).toContain('Non-scratch orgs');
+    expect(terminalText).toContain(`${devHubUserName}`);
+    expect(terminalText).toContain(`${devHubAliasName}`);
```
CristiCanizales, line 54: why are we using \`${}\` in all these above and not just the variables?

CristiCanizales (commented):
This looks great Ritam! Thank you for doing this 😄 Just a few minor comments/questions, plus the cla needs to be signed and the first commit is blocking that. As suggested over slack, to get pass through that you can squash merge all these changes to a new branch and then create a new PR 👍

test/setup/anInitialSetUp.e2e.ts
RitamAgrawal: Thanks for pointing this out. It should be changed to variables. Using ${} was one of many iterations I tried.

PR #48, opened 2023-08-30
Fixes for Apex LSP on GHA

CristiCanizales:
After upgrade of Apex LSP we need to make some adjustments to e2e (java 8 --> java 11)  so the LSP successfully activates on remote. 

@W-14016102@


test/specs/apexReplayDebugger.e2e.ts
daphne-sfdc, line 1: Line 106: The variable "textEditor" is not used.

test/specs/apexReplayDebugger.e2e.ts
```diff
@@ -122,6 +134,11 @@ describe('Apex Replay Debugger', async () => {
     );
     expect(successNotificationWasFound).toBe(true);
 
+    // Close finder tool
+    await browser.keys(['Escape']);
```
daphne-sfdc, line 138: Why did you put in three Escapes instead of just one?  Doesn't it take only one Escape to close the finder tool?  Or are you putting in extra escapes in case the first one doesn't do the job?

test/testSetup.ts
```diff
@@ -12,6 +12,8 @@ import util from 'util';
 import { DefaultTreeItem, InputBox, QuickOpenBox, Workbench } from 'wdio-vscode-service';
 import { EnvironmentSettings } from './environmentSettings';
 import * as utilities from './utilities';
+import vscode from 'wdio-vscode-service';
```
daphne-sfdc, line 15: This import isn't used in the file.

test/testSetup.ts
```diff
@@ -442,4 +440,88 @@ export class TestSetup {
...
+    if (!fs.existsSync(path.dirname(vscodeSettingsPath))) {
+      utilities.log(
+        `${this.testSuiteSuffixName} - ${path.dirname(vscodeSettingsPath)} does not exist`
+      );
+    }
+
+    const settings = fs.readFileSync(vscodeSettingsPath, 'utf8');
+
+    utilities.log(`${this.testSuiteSuffixName} - Workspace settings: ${JSON.stringify(settings)}`);
+  }
```
daphne-sfdc, line 480: This function is declared but never used.

test/testSetup.ts
```diff
@@ -243,7 +241,7 @@ export class TestSetup {
     );
   }
 
-  private async createDefaultScratchOrg(): Promise<void> {
+  public async createDefaultScratchOrg(): Promise<void> {
```
daphne-sfdc, line 244: Why is this changed from private to public?  The function is not called anywhere outside of testSetup.ts.

daphne-sfdc (changes requested):

test/specs/apexLsp.e2e.ts
```diff
@@ -36,6 +39,27 @@ describe('Apex LSP', async () => {
...
+      workbench,
+      'Editor Language Status'
+    );
+    await statusBar.click();
+    expect(await statusBar.getAttribute('aria-label')).toContain('Indexing complete');
+
+    // Get output text from the LSP
+    const outputViewText = await utilities.getOutputViewText('Apex Language Server');
+    utilities.log('Output view text');
+    utilities.log(outputViewText);
```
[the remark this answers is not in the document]
CristiCanizales: Yes, I just want to have that info logged for future reference

test/specs/apexReplayDebugger.e2e.ts
CristiCanizales: First one gets rid of the notification, second one closes the finder tool and the third one unselects the text previously selected

test/testSetup.ts
CristiCanizales: I used it somewhere when I was trying some things and I guess I forgot to change it back to private. Good catch, thanks!

PR #50, opened 2023-09-27
Visualforce E2E Test - zoom out to make Visualforce extension visible in running extensions

daphne-sfdc:
Previously, the Visualforce E2E test sometimes failed in the "Verify Extension is Running" step, which checks that the Visualforce extension is present in "Developer: Show Running Extensions".  The Visualforce extension is in fact started up but hidden at the bottom of the list and only accessible using the scroll bar.  To fix the problem, I added four "View: Zoom Out" lines before checking for the Visualforce extension in the list of running extensions so that all the running extensions are visible during the check.


CristiCanizales (approved):
Successful run: [LSP End to End Tests](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/6331065973)

PR #51, opened 2023-09-27
Improve consistency of notification checking in E2E tests

daphne-sfdc:
Our E2E tests rely on notification checking to make sure that commands are successfully run.  This PR improves the consistency of the notification checking process so that E2E tests don't frequently fail during this step.

### Original Behavior:
In every run of all E2E tests, there were 1-3 jobs that would fail during notification checking because of the following reasons:

1. The notification is fading away / already disappeared before the check was conducted
2. The command is still running and the success notification did not appear yet

This is happening because the amount of time it takes for a command to finish running fluctuates between different test runs and different OS.  It's difficult to catch the exact timing of the notifications' presence.

### Improved Behavior:
All runs of E2E tests stop failing during the notifications checking step of the workflow.

### Key Changes Made:
1. Replace all calls to notificationIsPresent() with notificationIsPresentWithTimeout().  This allows for repeated checking for the presence of the notification within a given period of time instead of just doing one check.
2. Remove all calls to waitForNotificationToGoAway() that are placed directly before calls to notificationIsPresent() because they are no longer needed.
3. Replace all calls to attemptToFindNotification() with notificationIsPresentWithTimeout().  Now, searching for notifications takes place during a set amount of time instead of a set number of tries.
4. Increase the timeout of calls to notificationIsPresentIWithTimeout() from 5 minutes to 10 minutes.  This ensures that all commands have enough time to finish running before their success notifications are being checked.

### Successful Run:
- [492](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/6330709049) (The second and third attempts were needed because some tests failed during setup)

@W-14096102@

### DIFF
--- a/test/specs/apexReplayDebugger.e2e.ts
+++ b/test/specs/apexReplayDebugger.e2e.ts
@@ -33,7 +33,7 @@ describe('Apex Replay Debugger', async () => {
     const successPushNotificationWasFound = await utilities.notificationIsPresentWithTimeout(
       workbench,
       'SFDX: Push Source to Default Org and Override Conflicts successfully ran',
-      utilities.TEN_MINUTES
+      utilities.FIFTEEN_MINUTES
     );
     expect(successPushNotificationWasFound).toBe(true);
   });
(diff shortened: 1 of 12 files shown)


CristiCanizales (approved):
This looks good to me! Thanks for this refactor
:shipit: 

Run: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/6332338275

PR #54, opened 2023-10-17
Fix Common Setup Errors in E2E Tests

daphne-sfdc:
This PR contains fixes for the most common setup errors that are causing E2E tests to fail.  Here is what got fixed:

1. **Problem**: "element not interactable" when running "SFDX: Create Project" to start each E2E test because the CLI Integration extension is not yet loaded.
**Solution**: Run "SFDX: Create Project" to trigger the CLI Integration extension to load, then hit Escape and check every 10 seconds that the extension is actually loaded.  After the extension is activated, run "SFDX: Create Project" again to completion.

2. **Problem**: "Quick Pick Item SFDX: ______ was not found" after running "Developer: Reload Window".
**Solution**: Every time "Developer: Reload Window" is run, wait for all extensions to start up and confirm that they are activated before they are called.

3. **Problem**: When switching to a different tab the Editor View, the tab is sometimes hidden to the right side because the VSCode window is too small.
**Solution**: Use "Go to File" in the command palette to open the file or switch to that tab instead of searching the visible tabs for the file name.

4. **Problem**: "element not interactable" when trying to click the 'Clear Output' button in the Output panel.
**Solution**: Use the command palette "View: Clear Output" instead of clicking the 'Clear Output' button.

5. **Problem**: Cannot click the "Prefer Deploy on Save" button in Settings because it is hidden by the Output panel when the VSCode window is too small.
**Solution**: Run "View: Close Panel" to make the setting visible on the page before clicking the checkbox.

This PR also contains a new helper function in extensionUtils.ts called verifyAllExtensionsAreRunning().  This helper function is used in testSetup.ts, after "Developer: Reload Window" (see Problem 2 above), and in a refactor of AnInitialSuite.e2e.ts to improve the verification of running extensions.

@W-14202398@

### DIFF
--- a/test/specs/anInitialSuite.e2e.ts
+++ b/test/specs/anInitialSuite.e2e.ts
@@ -103,30 +103,7 @@ describe('An Initial Suite', async () => {
   });
 
   step('Verify our extensions are loaded after creating an SFDX project', async () => {
-    const workbench = await (await browser.getWorkbench()).wait();
-    await utilities.runCommandFromCommandPrompt(workbench, 'Developer: Show Running Extensions', 3);
-    await utilities.enableLwcExtension();
-    // Close panel and clear all notifications so all extensions can be seen
-    await utilities.runCommandFromCommandPrompt(workbench, 'View: Close Panel', 1);
-    await utilities.runCommandFromCommandPrompt(
-      workbench,
-      'Notifications: Clear All Notifications',
-      1
-    );
-
-    let matchesFound = 0;
-    const extensionNameDivs = await $$('div.name');
-    for (const extensionNameDiv of extensionNameDivs) {
-      const text = await extensionNameDiv.getText();
-
-      if (text.startsWith('salesforce.salesforcedx-vscode-')) {
-        matchesFound++;
-        utilities.log(`AnInitialSuite - extension ${text} is loaded`);
-      }
-    }
-
-    // 7 are in the list, but sometimes only 6 are visible.
-    expect(matchesFound).toBeGreaterThanOrEqual(6);
+    testSetup.verifyAllExtensionsAreRunning();
   });
 
   step('Verify that SFDX commands are present after an SFDX project has been created', async () => {
--- a/test/specs/anInitialSuite.e2e.ts
+++ b/test/specs/anInitialSuite.e2e.ts
@@ -114,9 +114,6 @@ describe('An Initial Suite', async () => {
     const workbench = await (await browser.getWorkbench()).wait();
     const prompt = await utilities.openCommandPromptWithCommand(workbench, 'SFDX:');
     const quickPicks = await prompt.getQuickPicks();
-    // throw error for screenshot
-    let x = 0;
-    expect(x).toBe(2);
     const commands: string[] = [];
     for (const quickPick of quickPicks) {
       commands.push(await quickPick.getLabel());
--- a/test/specs/anInitialSuite.e2e.ts
+++ b/test/specs/anInitialSuite.e2e.ts
@@ -106,6 +106,8 @@ describe('An Initial Suite', async () => {
     utilities.verifyAllExtensionsAreRunning();
     browser.keys(['Escape']);
     await utilities.pause(1);
+    browser.keys(['Escape']);
+    await utilities.pause(1);
   });
 
   step('Verify that SFDX commands are present after an SFDX project has been created', async () => {
--- a/test/specs/anInitialSuite.e2e.ts
+++ b/test/specs/anInitialSuite.e2e.ts
@@ -103,17 +103,13 @@ describe('An Initial Suite', async () => {
   });
 
   step('Verify our extensions are loaded after creating an SFDX project', async () => {
-    utilities.log('1');
     testSetup.verifyAllExtensionsAreRunning();
-    utilities.log('2');
+    browser.keys(['Escape']);
   });
 
   step('Verify that SFDX commands are present after an SFDX project has been created', async () => {
     const workbench = await (await browser.getWorkbench()).wait();
     const prompt = await utilities.openCommandPromptWithCommand(workbench, 'SFDX:');
-    // throw error to see screenshot - remove this later
-    let x = 0;
-    expect(x).toBe(5);
     const quickPicks = await prompt.getQuickPicks();
     const commands: string[] = [];
     for (const quickPick of quickPicks) {
--- a/test/specs/anInitialSuite.e2e.ts
+++ b/test/specs/anInitialSuite.e2e.ts
@@ -103,7 +103,7 @@ describe('An Initial Suite', async () => {
   });
 
   step('Verify our extensions are loaded after creating an SFDX project', async () => {
-    testSetup.verifyAllExtensionsAreRunning();
+    utilities.verifyAllExtensionsAreRunning();
     browser.keys(['Escape']);
     await utilities.pause(1);
   });
--- a/test/specs/anInitialSuite.e2e.ts
+++ b/test/specs/anInitialSuite.e2e.ts
@@ -105,6 +105,7 @@ describe('An Initial Suite', async () => {
   step('Verify our extensions are loaded after creating an SFDX project', async () => {
     testSetup.verifyAllExtensionsAreRunning();
     browser.keys(['Escape']);
+    await utilities.pause(1);
   });
 
   step('Verify that SFDX commands are present after an SFDX project has been created', async () => {
--- a/test/specs/anInitialSuite.e2e.ts
+++ b/test/specs/anInitialSuite.e2e.ts
@@ -112,6 +112,9 @@ describe('An Initial Suite', async () => {
     const workbench = await (await browser.getWorkbench()).wait();
     const prompt = await utilities.openCommandPromptWithCommand(workbench, 'SFDX:');
     const quickPicks = await prompt.getQuickPicks();
+    // throw error for screenshot
+    let x = 0;
+    expect(x).toBe(2);
     const commands: string[] = [];
     for (const quickPick of quickPicks) {
       commands.push(await quickPick.getLabel());
--- a/test/specs/anInitialSuite.e2e.ts
+++ b/test/specs/anInitialSuite.e2e.ts
@@ -103,12 +103,17 @@ describe('An Initial Suite', async () => {
   });
 
   step('Verify our extensions are loaded after creating an SFDX project', async () => {
+    utilities.log('1');
     testSetup.verifyAllExtensionsAreRunning();
+    utilities.log('2');
   });
 
   step('Verify that SFDX commands are present after an SFDX project has been created', async () => {
     const workbench = await (await browser.getWorkbench()).wait();
     const prompt = await utilities.openCommandPromptWithCommand(workbench, 'SFDX:');
+    // throw error to see screenshot - remove this later
+    let x = 0;
+    expect(x).toBe(5);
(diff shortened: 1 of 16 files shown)


test/utilities/apexUtils.ts
```diff
@@ -116,8 +117,8 @@ export async function createAnonymousApexFile(): Promise<void> {
 
   // Set the name of the new Anonymous Apex file
   await inputBox.setText('Anonymous.apex');
-  await inputBox.confirm();
-  await inputBox.confirm();
+  await browser.keys(['Enter']);
+  await browser.keys(['Enter']);
```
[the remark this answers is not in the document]
daphne-sfdc: It's quite possible that inputBox.confirm() is flaky - it's actually still causing trouble now in createApexClass() in my last several runs of "Apex End to End Tests", even when I add additional pauses.  But browser.keys() was also acting flaky there (I switched from inputBox.confirm() to browser.keys() and back to inputBox.confirm()) so I'm not sure whether this change is better or worse.  At least this change seems to fix the problem for anonymous Apex classes.

test/specs/deployAndRetrieve.e2e.ts
```diff
@@ -290,6 +310,12 @@ describe('Deploy and Retrieve', async () => {
     );
     await browser.keys(['prefer deploy']);
 
+    try {
+      await utilities.runCommandFromCommandPrompt(workbench, 'View: Close Panel', 2);
+    }
+    catch {
+      utilities.log('Panel is already closed');
+    }
```
daphne-sfdc, line 318: This was added because the setting "Prefer Deploy on Save" is hidden by the Output panel when the VSCode window is too small.

test/testSetup.ts
```diff
@@ -82,6 +83,7 @@ export class TestSetup {
       3
     );
     await browser.keys(['Window: Command Center']);
+    await utilities.pause(3);
```
daphne-sfdc, line 86: Pause is needed here to give the settings time to load after searching.  Prevents "element not interactable" error when the setting is not visible on the page.

test/specs/apexReplayDebugger.e2e.ts
```diff
@@ -59,6 +59,7 @@ describe('Apex Replay Debugger', async () => {
     // Reloading the window forces the extensions to be reloaded and this seems to fix
     // the issue.
     await utilities.runCommandFromCommandPrompt(workbench, 'Developer: Reload Window', 30);
+    await testSetup.verifyAllExtensionsAreRunning();
```
daphne-sfdc, line 62: After reloading the VSCode window, the Salesforce extensions need time to activate before they are used.  Adding this line verifies that all extensions are activated after the reload.

test/specs/deployAndRetrieve.e2e.ts
```diff
@@ -88,10 +88,14 @@ describe('Deploy and Retrieve', async () => {
   });
 
   step('Deploy with SFDX: Deploy This Source to Org - ST enabled', async () => {
+    const workbench = await (await browser.getWorkbench()).wait();
     // Clear the Output view first.
     const outputView = await utilities.openOutputView();
-    await outputView.clearText();
-    const workbench = await (await browser.getWorkbench()).wait();
+    await utilities.runCommandFromCommandPrompt(workbench, 'View: Clear Output', 2);
```
daphne-sfdc, line 94: The "Clear Output" button in the Output panel often experiences flaky behavior - there is an "element not interactable" error if the button cannot be clicked.  Clicking the button can be replaced by running "View: Clear Output" in the command palette, which is much more stable.

test/testSetup.ts
```diff
@@ -165,6 +185,83 @@ export class TestSetup {
     utilities.log('');
   }
 
+  public async verifyAllExtensionsAreRunning(): Promise<void> {
```
daphne-sfdc, line 188: A new function that makes sure all Salesforce extensions are present and activated.  Called at the beginning of the workflow during setup and every time the E2E test runs "Developer: Reload Window".

test/testSetup.ts
```diff
@@ -165,6 +185,83 @@ export class TestSetup {
...
+      'salesforcedx-vscode-visualforce'
+    );
+    expect(vfExtensionWasFound).toBe(true);
+
+    // // Verify LWC extension is present and running
+    // const lwcExtensionWasFound = await utilities.findExtensionInRunningExtensionsList(
+    //   workbench,
+    //   'salesforcedx-vscode-lwc'
+    // );
+    // expect(lwcExtensionWasFound).toBe(true);
```
daphne-sfdc, line 255: The LWC extension does not start up properly on GHA, so it is omitted from the checks.  Only the LWC LSP E2E test uses this extension, so it is okay to not load it for all other E2E tests.

test/utilities/extensionUtils.ts
```diff
@@ -20,13 +20,21 @@ export async function findExtensionInRunningExtensionsList(
...
   // Close the panel and clear notifications so we can see as many of the running extensions as we can.
-  await runCommandFromCommandPrompt(workbench, 'View: Close Panel', 1);
+  try {
+    await runCommandFromCommandPrompt(workbench, 'View: Close Panel', 1);
+  }
+  catch {
+    // Close the command prompt by hitting the Escape key
+    await browser.keys(['Escape']);
+    log('No panel to close - command not found');
+  }
```
daphne-sfdc, line 30: Try/catch block is needed because "View: Close Panel" is not available in the command palette if the Output panel is already closed.

test/testSetup.ts
```diff
@@ -123,15 +125,33 @@ export class TestSetup {
...
+    }
+    while (coreExtensionWasFound === false);
+    utilities.log(`${this.testSuiteSuffixName} - Ready to create the standard project`);
+
+    this.prompt = await utilities.runCommandFromCommandPrompt(workbench, 'SFDX: Create Project', 5);
     // Selecting "SFDX: Create Project" causes the extension to be loaded, and this takes a while.
 
     // Select the "Standard" project type.
-    await utilities.selectQuickPickWithText(this.prompt, 'Standard');
+    await utilities.selectQuickPickItem(this.prompt, 'Standard');
```
daphne-sfdc, line 154: This is the only place where selectQuickPickWithText() is used.  The only difference between selectQuickPickWithText() and selectQuickPickItem() is that we can see the word "Standard" being typed out before it is selected.  Making this replacement therefore contains very minimal change in behavior, and E2E tests are still able to be run successfully.

test/specs/pushAndPull.e2e.ts
```diff
@@ -24,7 +24,7 @@ describe('Push and Pull', async () => {
 
   step('Set up the testing environment', async () => {
     testSetup = new TestSetup('PushAndPull', false);
-    await testSetup.setUp('Enterprise');
+    await testSetup.setUp();
```
daphne-sfdc, line 27: I'm not sure why PushAndPull used to create Enterprise scratch orgs.  All the other E2E tests create Developer scratch orgs (which is the default), so I changed PushAndPull to also do the same.  The test continues to run successfully after making this switch.

test/specs/apexLsp.e2e.ts
```diff
@@ -64,6 +64,10 @@ describe('Apex LSP', async () => {
     utilities.log(`${testSetup.testSuiteSuffixName} - Go to Definition`);
     // Get open text editor
     const workbench = await (await browser.getWorkbench()).wait();
+    const inputBox = await utilities.runCommandFromCommandPrompt(workbench, 'Go to File...', 1);
+    await inputBox.setText('ExampleClassTest.cls');
+    await inputBox.confirm();
+    await utilities.pause(1);
```
daphne-sfdc, line 70: This is added to every place where the E2E test switches from one tab in the editor view to another tab. It's needed because when the VSCode window is too small, some tabs are hidden to the right side and require scrolling to access. Using "Go to File" is a replacement for scrolling, which WDIO does not support.

diyer:
for problem 1, what is the determining factor to wait for 10 seconds? can there be some setup method that can be called that loads the extenstion before create project?

test/testSetup.ts
```diff
@@ -165,6 +185,83 @@ export class TestSetup {
...
+    //   'salesforcedx-vscode-lwc'
+    // );
+    // expect(lwcExtensionWasFound).toBe(true);
+
+    await utilities.pause(2);
+    await utilities.runCommandFromCommandPrompt(workbench, 'View: Zoom In', 2);
+    await utilities.runCommandFromCommandPrompt(workbench, 'View: Zoom In', 2);
+    await utilities.runCommandFromCommandPrompt(workbench, 'View: Zoom In', 2);
+    await utilities.runCommandFromCommandPrompt(workbench, 'View: Zoom In', 2);
+    await utilities.pause(2);
```
CristiCanizales, line 262: Same here. Do we need this pause? At this point extensions have been verified already

CristiCanizales (commented):
Great work on this Daphne! Seems like we'll be way closer on relying on e2e on gha with this 🥳 Just a few comments/questions

CristiCanizales:
> for problem 1, what is the determining factor to wait for 10 seconds? can there be some setup method that can be called that loads the extenstion before create project?

I guess it could be 5 seconds and it would maybe iterate more times; the thing is that extensions take some time to load, even longer on windows. The do while she implemented would be that method that keeps checking if the CLI extensions is already there

test/testSetup.ts
daphne-sfdc: This one is probably unnecessary.

test/testSetup.ts
CristiCanizales: Can we move this to extensionUtils? seems like a better place to be as it's going to be used in anInitialSuite too

daphne-sfdc: It's moved now!

test/specs/anInitialSuite.e2e.ts
```diff
@@ -103,30 +103,11 @@ describe('An Initial Suite', async () => {
...
-      }
-    }
-
-    // 7 are in the list, but sometimes only 6 are visible.
-    expect(matchesFound).toBeGreaterThanOrEqual(6);
+    utilities.verifyAllExtensionsAreRunning();
+    browser.keys(['Escape']);
+    await utilities.pause(1);
+    browser.keys(['Escape']);
+    await utilities.pause(1);
```
daphne-sfdc, line 110: verifyAllExtensionsAreRunning() in extensionUtils.ts is a newly created utility function.  It checks if all the Salesforce extensions are present and running, and a call to that function can replace all the code in this step.

test/utilities/extensionUtils.ts
```diff
@@ -41,3 +42,78 @@ export async function findExtensionInRunningExtensionsList(
 
   return extensionWasFound;
 }
+
+export async function verifyAllExtensionsAreRunning(): Promise<void> {
```
daphne-sfdc, line 46: This verifyAllExtensionsAreRunning() function started out as a step in setup, but its use cases grew beyond setup to "Developer: Reload Window" and AnInitialSuite.  So it's now moved here to extensionUtils.ts per @CristiCanizales's suggestion.

CristiCanizales:
QE run: [End to End Tests](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/6604470323)

CristiCanizales (approved):
QE notes:
The QE run had a very high success rate, more than before and none of the errors specified in this PR were thrown hence approving this PR.
Thank you Daphne! :awesome:

PR #55, opened 2023-10-18
Move lwcLsp.e2e.ts back to specs folder

daphne-sfdc:
lwcLsp.e2e.ts was previously moved to the TBD folder because it could not be run against the source code on GHA.  Now that the LWC extension can be installed via VSIX, lwcLsp.e2e.ts can be moved back to the specs folder and utilized for testing.


CristiCanizales (approved):
Thanks :)

🚢

PR #56, opened 2023-10-18
chore: refactor findExtensionInRunningExtensionsList

CristiCanizales:
This is just a refactor of `findExtensionInRunningExtensionsList()` in order to check for the right id in the html and not what's displayed


daphne-sfdc (approved):
Approved! 🎉 Thanks for the fix!

PR #65, opened 2024-01-10
E2E test improvements - (1) fix for "coordinates is not iterable" error in LWC LSP (2) SFDX -> SF

daphne-sfdc:
This PR contains two main improvements to the E2E tests:
1. Previously, we were seeing a "coordinates is not iterable" error in about 80% of our Ubuntu test runs of the LWC LSP E2E test.  This PR replaces the `.typeTextAt()` function with `browser.keys` in order to work around that issue.
2. In our code, several CLI commands are called using `exec()` to be run directly in the Terminal.  The following CLI commands are changed from SFDX-style to SF-style:
a. `auth:sfdxurl:store` -> `org:login:sfdx-url`
b. `force:user:create` -> `org:create:user`
c. `force:org:delete` -> `org:delete:scratch`
d. `force:org:display` -> `org:display`
e. `force:org:list` -> `org:list`
f. `force:org:create` -> `org:create:scratch`

Passing E2E test run: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/7474955893

@W-14791862@
@W-14791836@

### DIFF
--- a/test/specs/lwcLsp.e2e.ts
+++ b/test/specs/lwcLsp.e2e.ts
@@ -96,7 +96,6 @@ describe('LWC LSP', async () => {
     await browser.keys([CMD_KEY, 'f']);
     await utilities.pause(1);
     await browser.keys(['div']);
-    await browser.keys(['Enter']);
     await browser.keys(['Escape']);
     await browser.keys(['ArrowRight']);
     await utilities.pause(1);
--- a/test/specs/lwcLsp.e2e.ts
+++ b/test/specs/lwcLsp.e2e.ts
@@ -70,9 +70,7 @@ describe('LWC LSP', async () => {
     // Move cursor to the middle of "greeting"
     await browser.keys([CMD_KEY, 'f']);
     await utilities.pause(1);
-    await browser.keys(['greeting']);
-    await browser.keys(['Escape']);
-    await browser.keys(['ArrowRight', 'ArrowLeft', 'ArrowLeft']);
+    await browser.keys(['greeting','Escape','ArrowRight', 'ArrowLeft', 'ArrowLeft']);
     await utilities.pause(1);
 
     // Go to definition through F12
--- a/test/specs/lwcLsp.e2e.ts
+++ b/test/specs/lwcLsp.e2e.ts
@@ -70,7 +70,7 @@ describe('LWC LSP', async () => {
     // Move cursor to the middle of "greeting"
     await browser.keys([CMD_KEY, 'f']);
     await utilities.pause(1);
-    await browser.keys(['greeting','Escape','ArrowRight', 'ArrowLeft', 'ArrowLeft']);
+    await browser.keys(['greeting', 'Escape', 'ArrowRight', 'ArrowLeft', 'ArrowLeft']);
     await utilities.pause(1);
 
     // Go to definition through F12
@@ -93,9 +93,7 @@ describe('LWC LSP', async () => {
     // Move cursor to right after the first 'div' tag and type ' lwc'
     await browser.keys([CMD_KEY, 'f']);
     await utilities.pause(1);
-    await browser.keys(['div']);
-    await browser.keys(['Escape']);
-    await browser.keys(['ArrowRight']);
+    await browser.keys(['div', 'Escape', 'ArrowRight']);
     await utilities.pause(1);
     await browser.keys([' lwc']);
     await utilities.pause(2);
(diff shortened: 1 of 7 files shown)


test/specs/lwcLsp.e2e.ts
```diff
@@ -55,7 +55,7 @@ describe('LWC LSP', async () => {
     await utilities.pause(1);
 
     // Verify 'Go to definition' took us to the definition file
-    const editorView = await workbench.getEditorView();
+    const editorView = workbench.getEditorView();
```
CristiCanizales, line 58: nice catch! Could you also remove that await in other places this function is called?

CristiCanizales:
@daphne-sfdc could you also please create a WI to track this work and tag it in the PR description?

test/testSetup.ts
```diff
@@ -325,11 +325,11 @@ export class TestSetup {
     const startDate = Date.now();
     const durationDays = 1;
 
-    utilities.log(`${this.testSuiteSuffixName} - calling "sfdx force:org:create"...`);
+    utilities.log(`${this.testSuiteSuffixName} - calling "sfdx org:create:scratch"...`);
     const sfdxForceOrgCreateResult = await exec(
-      `sfdx force:org:create -f ${definitionFile} --setalias ${this.scratchOrgAliasName} --durationdays ${durationDays} --setdefaultusername --json --loglevel fatal`
+      `sfdx org:create:scratch --edition developer --definition-file ${definitionFile} --alias ${this.scratchOrgAliasName} --duration-days ${durationDays} --json`
```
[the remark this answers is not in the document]
daphne-sfdc: Thanks for catching that and providing the replacement command! 🙂

test/specs/lwcLsp.e2e.ts
```diff
@@ -84,7 +91,15 @@ describe('LWC LSP', async () => {
     // Get open text editor
     const workbench = await browser.getWorkbench();
     const textEditor = await utilities.getTextEditor(workbench, 'lwc1.html');
-    await textEditor.typeTextAt(3, 7, ' lwc');
+
+    // Move cursor to right after the first 'div' tag and type ' lwc'
+    await browser.keys([CMD_KEY, 'f']);
+    await utilities.pause(1);
+    await browser.keys(['div']);
```
peternhale, line 98: Same here?

test/testSetup.ts
```diff
@@ -60,7 +60,7 @@ export class TestSetup {
       // await utilities.executeCommand(workbench, `sfdx force:org:delete -u ${this.scratchOrgAliasName} --noprompt`);
 
       // The Terminal view can be a bit unreliable, so directly call exec() instead:
-      await exec(`sfdx force:org:delete -u ${this.scratchOrgAliasName} --noprompt`);
+      await exec(`sfdx org:delete:scratch --target-org ${this.scratchOrgAliasName} --no-prompt`);
```
peternhale, line 63: If we don't want to use terminal view, how about cleaning up the commented code above?

peternhale: And why are we holding on to code that no longer works?
Is there a work item to look into why so it can be re-enabled at some point?

peternhale (approved):
@daphne-sfdc Looks good, just a few questions and one possible change.

test/specs/lwcLsp.e2e.ts
daphne-sfdc: This one cannot be combined because of the pause on Line 97 - that was an intentional pause to wait for the Find box to open.

test/testSetup.ts
daphne-sfdc: Typically for E2E tests, if one method doesn't work we just find a workaround.  Since using `exec()` is a good workaround, we are likely not switching back to using Terminal view (as testing the CLI is not the purpose of the E2E tests so it doesn't matter how we run the commands).  I can clean up the commented code.

daphne-sfdc: I see that one block of commented code has the functionality of deleting the folder where the project was located during the tear down process.  We don't need that anymore because each E2E test is now run individually on a separate VM, so it doesn't matter if we do that cleanup.

test/specs/lwcLsp.e2e.ts
peternhale: @daphne-sfdc I am confused. We have
await browser.keys(['div']);
await browser.keys(['Escape']);
await browser.keys(['ArrowRight']);
await utilities.pause(1);

Why cannot this become?
await browser.keys(['div','Escape','ArrowRight']);
await utilities.pause(1);

daphne-sfdc: Oh sorry!  Yes you're right, Github only previewed the first four lines, so I only saw `await browser.keys(['div']);` and didn't realize what you were referring to.

CristiCanizales (approved):
LGTM!

:shipit: 

Test run: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/7490318508

PR #67, opened 2024-01-30
fix: eliminate 'unknown error: unhandled inspector error: {"code":-32000,"message":"No node with given id found"}' flapper

daphne-sfdc:
This flapper has been around for a few months, and it was mostly present in **ApexLsp.e2e.ts** during the **Verify Extension is Running** step.  The cause of this flapper was in **extensionUtils.ts** in the `findExtensionInRunningExtensionsList()` function - after the extension is found, the for loop kept going and searching for the extension rather than breaking out of the loop, which caused a stale element and this error message.

What was fixed: Add a `break` in the for loop after the extension is found.

@W-14498987@

Passing E2E test runs:

1. Apex E2E Tests #84: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/7704042747
2. Apex E2E Tests #85: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/7704043867
3. Apex E2E Tests #86: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/7704045623
4. E2E Tests #768: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/7714238818
5. E2E Tests #770: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/7714244040

### DIFF
--- a/test/specs/apexLsp.e2e.ts
+++ b/test/specs/apexLsp.e2e.ts
@@ -4,7 +4,7 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import { step } from 'mocha-steps';
+import { step, xstep } from 'mocha-steps';
 import { TestSetup } from '../testSetup';
 import * as utilities from '../utilities';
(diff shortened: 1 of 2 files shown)


CristiCanizales (approved):
LGTM! Thanks for digging into this and finding a fix 🥳 
:shipit:

PR #72, opened 2024-04-08
fix: Aura E2E Test runs on VSIXs

daphne-sfdc:
This PR contains fixes for flappers in the Aura E2E test that prevent it from running successfully on VSIXs.

What got fixed:
1. The Aura LSP indexes existing Aura files when the Aura extension is activated, but the Aura component that we use for Go to Definition did not get indexed because it was created after the indexing happened.  I added a reload after the Aura component is created so that the Aura LSP would restart and redo the indexing, capturing the new component.
2. The `textEditor.getCoordinates()` function call in Go to Definition and the `textEditor.typeTextAt()` function call in Autocompletion cause a `coordinates is not iterable` error when the E2E test is run on Ubuntu.  This PR contains workarounds using `browser.keys()` to recreate the same functionality and get around the issue.

Passing E2E test runs:
1. https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/8575950666 ✅ 
3. https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/8576151245 ✅ 

@W-14385264@

### DIFF
--- a/test/specs/auraLsp.e2e.ts
+++ b/test/specs/auraLsp.e2e.ts
@@ -74,7 +74,7 @@ describe('Aura LSP', async () => {
     expect(definition[1]).toBe(27);
   });
 
-  xstep('Autocompletion', async () => {
+  step('Autocompletion', async () => {
     utilities.log(`${testSetup.testSuiteSuffixName} - Autocompletion`);
     // Get open text editor
     const workbench = await (await browser.getWorkbench()).wait();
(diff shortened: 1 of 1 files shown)


CristiCanizales (approved):
Looks good! Thank you for findind a way to have this test back 🥳 
E2E Test run: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/8603236768 ✅

PR #77, opened 2024-05-10
feat: launch ard from current test class

CristiCanizales:
Implement test for SFDX: Launch Apex Replay Debugger from Current File
[skip-validate-pr]

Test run: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/9025691296/job/24801808352 ✅

### DIFF
--- a/test/specs/apexReplayDebugger.e2e.ts
+++ b/test/specs/apexReplayDebugger.e2e.ts
@@ -237,6 +237,13 @@ describe('Apex Replay Debugger', async () => {
     await utilities.pause(1);
     await browser.keys(['F5']);
     await utilities.pause(1);
+
+    const successNotificationWasFound = await utilities.notificationIsPresentWithTimeout(
+      workbench,
+      'Debug Test(s) successfully ran',
+      utilities.TEN_MINUTES
+    );
+    expect(successNotificationWasFound).toBe(true);
   });
 
   step('Run the Anonymous Apex Debugger using the Command Palette', async () => {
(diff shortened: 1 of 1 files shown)


test/specs/apexReplayDebugger.e2e.ts
```diff
@@ -222,6 +222,23 @@ describe('Apex Replay Debugger', async () => {
     await utilities.pause(1);
   });
 
+  step('SFDX: Launch Apex Replay Debugger with Current File - test class', async () => {
```
peternhale, line 225: @CristiCanizales there are no assertions made in this step. Is success making it through the code in this test. What if all the UI interactions succeed, but the launch actually fails?

peternhale (changes requested):

test/specs/apexReplayDebugger.e2e.ts
CristiCanizales: Right! I copied the previous step which throws no notif at all and didn't realize when launching from a test file we do get a notif that we can validate. Thanks! Test run: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/9032667289/job/24821299340

test/specs/apexReplayDebugger.e2e.ts
```diff
@@ -222,6 +222,30 @@ describe('Apex Replay Debugger', async () => {
...
+    const workbench = await (await browser.getWorkbench()).wait();
+    await utilities.getTextEditor(workbench, 'ExampleApexClassTest.cls');
+    await utilities.runCommandFromCommandPrompt(
+      workbench,
+      'SFDX: Launch Apex Replay Debugger with Current File',
+      3
+    );
+
+    // Continue with the debug session
+    await browser.keys(['F5']);
```
peternhale, line 236: Replace with constant 'CONTINUE'

peternhale (approved):

PR #85, opened 2024-07-01
fix: get apex lsp test running in esm branch

peternhale:
@W-16124376@


test/specs/runLwcTests.e2e.ts
```diff
@@ -47,12 +48,12 @@ describe('Run LWC Tests', async () => {
       await utilities.runCommandFromCommandPrompt(
         workbench,
         'SFDX: Run All Lightning Web Component Tests',
-        1
+        Duration.seconds(2)
```
daphne-sfdc, line 51: Why the change from 1 second to 2 seconds?  Is 1 second not long enough?

PR #87, opened 2024-07-05
fix: make e2e tests run past setup step on windows

CristiCanizales:
E2E tests were always failing with `Error: Quick pick item SFDX: Set a Default Org was not found` on windows. Tests can now run past that point and eventually (considering flappers like the rest of os's) make it to tear down.

@W-16164065@

[Test run](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/9812786679)

### DIFF
--- a/test/utilities/apexUtils.ts
+++ b/test/utilities/apexUtils.ts
@@ -6,7 +6,7 @@
  */
 
 import { TextEditor } from 'wdio-vscode-service';
-import { executeQuickPick, runCommandFromCommandPrompt } from './commandPrompt.ts';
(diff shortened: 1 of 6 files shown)


test/utilities/apexUtils.ts
daphne-sfdc, line 1: Line 111 - Can `const inputBox = await runCommandFromCommandPrompt(workbench, 'Create: New File...', 1);` use `executeQuickPick()` too?  That way the import of `runCommandFromCommandPrompt` can be removed.

daphne-sfdc (approved):

PR #88, opened 2024-07-05
chore: more factoring

peternhale:


test/utilities/extensionUtils.ts
```diff
@@ -24,7 +25,8 @@ export type ExtensionId =
   | 'salesforcedx-vscode-apex-replay-debugger'
   | 'salesforcedx-vscode-lightning'
   | 'salesforcedx-vscode-lwc'
-  | 'salesforcedx-vscode-visualforce';
+  | 'salesforcedx-vscode-visualforce'
+  | 'salesforce.salesforce-vscode-slds';
```
CristiCanizales, line 29: Why do we need to have this extension here? And at the same time removing the validation of the commands in anInitialSuite.e2e.ts?

test/specs/debugLwcTests.e2e.ts
```diff
@@ -14,7 +14,7 @@ import util from 'util';
 import { fail } from 'assert';
 
 import { Key } from 'webdriverio';
-const CMD_KEY = process.platform === 'darwin' ? Key.Command : Key.Control;
+const CMD_KEY = process.platform === 'darwin' ? Key.Command : Key.Control;import { Duration } from '@salesforce/kit';
```
CristiCanizales, line 17:
```suggestion
const CMD_KEY = process.platform === 'darwin' ? Key.Command : Key.Control;
import { Duration } from '@salesforce/kit';
```

test/specs/orgBrowser.e2e.ts
```diff
@@ -156,7 +169,7 @@ describe('Org Browser', async () => {
     utilities.log(`${testSetup.testSuiteSuffixName} - Retrieve and Open Source`);
     const myClassLabelEl = await utilities.findElementByText('div', 'aria-label', 'MyClass');
     myClassLabelEl.click();
-    utilities.pause(2);
+    utilities.pause(Duration.seconds(2));
```
CristiCanizales, line 172:
```suggestion
    await utilities.pause(Duration.seconds(2));
```

CristiCanizales (changes requested):
Looks great! Just some questions:

1. what is the benefit of using Duration.seconds(x) instead of x?
2. why do we include seconds in clearOutputView() in some places and not in others?

test/specs/manifestBuilder.e2e.ts
```diff
@@ -120,11 +121,11 @@ describe('Manifest Builder', async () => {
     const workbench = await browser.getWorkbench();
     await utilities.getTextEditor(workbench, 'manifest.xml');
     // Clear output before running the command
-    await utilities.runCommandFromCommandPrompt(workbench, 'View: Clear Output', 1);
+    await utilities.clearOutputView(Duration.seconds(2));
```
daphne-sfdc, line 124: It looks like `clearOutputView()` sometimes has a duration of 1 second and sometimes has a duration of 2 seconds.  Would it make sense to standardize it to 1 second throughout the code, especially since that's the default timeout?

test/specs/manifestBuilder.e2e.ts
```diff
@@ -120,11 +121,11 @@ describe('Manifest Builder', async () => {
...
     await utilities.runCommandFromCommandPrompt(
       workbench,
       'SFDX: Retrieve Source in Manifest from Org',
-      1
+      Duration.seconds(2)
```
daphne-sfdc, line 128: Why was this changed from 1 second to 2 seconds?  Is there a flapper that happens when it's 1 second?

test/utilities/extensionUtils.ts
daphne-sfdc: I removed the validation of the commands in anInitialSuite.e2e.ts because SLDS is installed as part of the Salesforce Extension Pack and we no longer install the packs in E2E.  In this case I agree with Cristi that Line 29 can be removed.

test/utilities/miscellaneous.ts
```diff
@@ -8,16 +8,18 @@
...
 import { executeQuickPick } from './commandPrompt.ts';
 import { notificationIsPresentWithTimeout } from './notifications.ts';
+import { Duration } from '@salesforce/kit';
 import path from 'path';
+import { PredicateWithTimeout } from './predicates.ts';
 
-export const FIVE_MINUTES = 5 * 60;
-export const TEN_MINUTES = 10 * 60;
+export const FIVE_MINUTES = Duration.minutes(5);
+export const TEN_MINUTES = Duration.minutes(10);
```
daphne-sfdc, line 19: 😀

test/utilities/outputView.ts
```diff
@@ -58,68 +52,59 @@ export async function attemptToFindOutputPanelText(
...
-    if (outputViewText.includes(searchString)) {
-      return outputViewText;
+  browser.waitUntil(
+    async () => {
+      const outputViewText = await getOutputViewText();
+      return outputViewText.includes(searchString);
+    },
+    {
+      timeout: 1000 * attempts, // Convert attempts to milliseconds assuming each attempt is 1 second
+      interval: 500, // Check every second
```
daphne-sfdc, line 62: Should the comment be "Check every half second" instead?

daphne-sfdc, line 62: Why is the interval 500 instead of 1000?

test/utilities/workbench.ts
```diff
@@ -43,24 +44,25 @@ export async function showExplorerView(): Promise<void> {
...
   await executeQuickPick('View: Reset Zoom', wait);
 }
 
-async function handlePredicateOrWait(predicateOrWait: PredicateWithTimeout | number, prompt: unknown) {
-  if (typeof predicateOrWait === 'number') {
-    if (predicateOrWait > 0) {
+async function handlePredicateOrWait(predicateOrWait: PredicateWithTimeout | Duration, prompt: unknown) {
+  log('handlePredicateOrWait')
+  if (isDuration(predicateOrWait)) {
+    if (predicateOrWait && predicateOrWait.milliseconds > 0) {
```
daphne-sfdc, line 62:
```suggestion
    if (predicateOrWait?.milliseconds > 0) {
```

test/utilities/extensionUtils.ts
peternhale: I'll remove the unneeded extension.

test/specs/manifestBuilder.e2e.ts
peternhale: I'll change it to 1 second

test/specs/manifestBuilder.e2e.ts
peternhale: here too

test/utilities/outputView.ts
peternhale: reduce overall wait time

peternhale: I'll fix the comment

test/utilities/workbench.ts
peternhale: The type assertion on line 62 guarantees predicateOrWait is not undefined, but is also of type Duration, so the safe navigation is not needed either.

peternhale:
> Looks great! Just some questions:
> 
> 1. what is the benefit of using Duration.seconds(x) instead of x?
> 2. why do we include seconds in clearOutputView() in some places and not in others?

1. We all know that the calls to sleep in pause require milliseconds, yet pause function's parameter is just a number that everyone should know that the unit is seconds. Changing the type to Duration takes away the need to know how the parameter is used within pause.  It also gives the test writer the ability to express the value in anyway they would like. Granted the most common use is using seconds, but now we can express the values in any time unit we would like. Duration.milliseconds(1000), Duration.seconds(1),  Duration.minutes(10). And when consumed, we can use the unit suitable for the need, `sleep(wait.milliseconds)`. Also, the change now allows us to use a unit smaller than a second, `pause(Duration.milliseconds(500))` without having to deal with unit conversions in the consuming function.
2. I believe I set a default duration of 1 second, so not providing uses the default.

test/specs/pushAndPull.e2e.ts
```diff
@@ -94,70 +97,67 @@ describe('Push and Pull', async () => {
...
+    await utilities.clearOutputView(Duration.seconds(2));
 
     // Now push
-    await utilities.runCommandFromCommandPrompt(workbench, 'SFDX: Push Source to Default Org', 5);
-
-    await verifyPushSuccess(workbench);
+    await utilities.executeQuickPick('SFDX: Push Source to Default Org', Duration.seconds(5));
 
     // Check the output.
-    await verifyPushAndPullOutputText(workbench, 'Push', 'to');
```
daphne-sfdc, line 127: Why don't we want to verify the output text after the push anymore?

test/specs/pushAndPull.e2e.ts
```diff
@@ -94,70 +97,67 @@ describe('Push and Pull', async () => {
...
     const textEditor = await utilities.getTextEditor(workbench, 'ExampleApexClass1.cls');
     await textEditor.setTextAtLine(3, '        // sample comment');
 
     // Push the file.
-    await utilities.runCommandFromCommandPrompt(workbench, 'SFDX: Push Source to Default Org', 5);
+    await utilities.executeQuickPick('SFDX: Push Source to Default Org', Duration.seconds(5));
 
     await verifyPushSuccess(workbench);
 
-    await verifyPushAndPullOutputText(workbench, 'Push', 'to');
```
daphne-sfdc, line 145: Same here - why is the validation removed?

test/specs/pushAndPull.e2e.ts
```diff
@@ -183,87 +183,78 @@ describe('Push and Pull', async () => {
...
+    await utilities.clearOutputView(Duration.seconds(2));
 
     // And pull again.
-    await utilities.runCommandFromCommandPrompt(workbench, 'SFDX: Pull Source from Default Org', 5);
-    await verifyPullSuccess(workbench);
-
+    await utilities.executeQuickPick('SFDX: Pull Source from Default Org', Duration.seconds(5));
     // Check the output.
-    outputPanelText = await verifyPushAndPullOutputText(workbench, 'Pull', 'from');
-    expect(outputPanelText).not.toContain('Created  Admin');
```
daphne-sfdc, line 211: And here - validation got removed.

test/specs/pushAndPull.e2e.ts
```diff
@@ -183,87 +183,78 @@ describe('Push and Pull', async () => {
...
     await textEditor.setTextAtLine(3, '        // sample comment for the pull test');
     // Don't save the file just yet.
 
     // Pull the file.
-    await utilities.runCommandFromCommandPrompt(workbench, 'SFDX: Pull Source from Default Org', 5);
-    await verifyPullSuccess(workbench);
-
+    await utilities.executeQuickPick('SFDX: Pull Source from Default Org', Duration.seconds(5));
     // Check the output.
-    await verifyPushAndPullOutputText(workbench, 'Pull', 'from');
```
daphne-sfdc, line 230: Also here - the validation is gone.

test/specs/pushAndPull.e2e.ts
```diff
@@ -183,87 +183,78 @@ describe('Push and Pull', async () => {
...
 
     // Now save the file.
     const textEditor = await utilities.getTextEditor(workbench, 'ExampleApexClass1.cls');
     await textEditor.save();
 
     // An now pull the changes.
-    await utilities.runCommandFromCommandPrompt(workbench, 'SFDX: Pull Source from Default Org', 5);
+    await utilities.executeQuickPick('SFDX: Pull Source from Default Org', Duration.seconds(5));
     await verifyPullSuccess(workbench);
-    await verifyPushAndPullOutputText(workbench, 'Pull', 'from');
```
daphne-sfdc, line 246: Here too for the missing validation.

test/specs/runEinsteinForDeveloper.e2e.ts
daphne-sfdc, line 1: Why is this file in our automation tests repo?  Doesn't E4D have their own repo with their E2E tests already?  (Or are we using it to be able to run their tests from our workflows?)

test/utilities/miscellaneous.ts
```diff
@@ -81,14 +91,14 @@ export async function createCommand(
   folder: string,
   extension: string
 ): Promise<string | undefined> {
-  const workbench = await (await browser.getWorkbench()).wait();
-  await executeQuickPick('View: Clear Output', 1);
-  const inputBox = await executeQuickPick(`SFDX: Create ${type}`, 1);
+  const workbench = await getWorkbench();
+  await executeQuickPick('View: Clear Output', Duration.seconds(1));
```
daphne-sfdc, line 95: Import and use the helper function from `outputView.ts`

```suggestion
  await clearOutputView();
```

test/environmentSettings.ts
```diff
@@ -35,7 +35,8 @@ export class EnvironmentSettings {
     // './specs/**/templates.e2e.ts',
     // './specs/**/trailApexReplayDebugger.e2e.ts',
     // './specs/**/visualforceLsp.e2e.ts',
-    // './specs/**/runEinsteinForDeveloper.e2e.ts'
+    // './specs/**/runEinsteinForDeveloper.e2e.ts',
```
daphne-sfdc, line 38: If we want to remove the `runEinsteinForDeveloper.e2e.ts` file from this repo, this line should also be removed.

test/specs/runEinsteinForDeveloper.e2e.ts
peternhale: I don't know but I will remove it.

test/utilities/extensionUtils.ts
```diff
@@ -12,6 +12,7 @@ import FastGlob from 'fast-glob';
 import { EnvironmentSettings } from '../environmentSettings.ts';
 import { exec } from 'child_process';
 import * as utilities from './index.ts';
+import { Duration } from '@salesforce/kit';
 
 export type ExtensionId =
```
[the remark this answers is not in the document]
peternhale: I recall discussions around writing simple, will our extension packs start, so maybe it should stay here until such time where we know the scope of the "will our extension packs start" tests.

test/specs/pushAndPull.e2e.ts
daphne-sfdc: Is this actually meant to be removed?  I see that this one is the only `verifyPushAndPullOutputText()` that is missing now.

daphne-sfdc (approved):
Approved! 🎉 

Thanks Pete for all the great refactoring work done in this PR! 😀 

E2E test run here: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/9944040002 ✅

PR #92, opened 2024-07-22
Modify E2E TestSetup so it will not fail if scratch org is not present

CristiCanizales:
@W-16262411@

### DIFF
--- a/test/testSetup.ts
+++ b/test/testSetup.ts
@@ -54,7 +54,11 @@ export class TestSetup {
 
   public async tearDown(): Promise<void> {
     await this.checkForUncaughtErrors();
-    await utilities.deleteScratchOrg(this.scratchOrgAliasName, this.reuseScratchOrg);
+    try {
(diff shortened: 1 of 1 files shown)


test/testSetup.ts
```diff
@@ -54,7 +54,11 @@ export class TestSetup {
 
   public async tearDown(): Promise<void> {
     await this.checkForUncaughtErrors();
-    await utilities.deleteScratchOrg(this.scratchOrgAliasName, this.reuseScratchOrg);
+    try {
+      await utilities.deleteScratchOrg(this.scratchOrgAliasName, this.reuseScratchOrg);
+    } catch (error) {
+      throw new Error(`Deleting scratch org failed with Error: ${(error as Error).message}`);
```
peternhale, line 60: @CristiCanizales why fail the test run because scratch org could not be deleted?

CristiCanizales: copy and paste error 🤡 I meant utilities.log

peternhale (approved):

PR #94, opened 2024-07-23
Cristi/delete scratch org info

CristiCanizales:
@W-16067588@

Test runs: 
- [End to End Tests](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/10143289683) ✅ 
- [E2E Test Suite - ESM](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/10143293454) ✅


test/testSetup.ts
```diff
@@ -24,6 +24,7 @@ export class TestSetup {
   public projectFolderPath: string | undefined;
   private prompt: QuickOpenBox | InputBox | undefined;
   public scratchOrgAliasName: string | undefined;
+  public scratchOrgId: string | undefined;
```
peternhale, line 27: How does this get set?

test/testSetup.ts
```diff
@@ -56,8 +57,26 @@ export class TestSetup {
...
     } catch (error) {
-      utilities.log(`Deleting scratch org failed with Error: ${(error as Error).message}`);
+      utilities.log(
+        `Deleting scratch org (or info) failed with Error: ${(error as Error).message}`
+      );
+    }
+  }
+
+  private async deleteScratchOrgInfo(): Promise<void> {
+    const sfDataDeleteRecord = await utilities.runCliCommand(
```
peternhale, line 69: This should only be done if `this.scratchOrgId` is not undefined.

test/testSetup.ts
peternhale: Why do we need this when we already know `scratchOrgAliasName`?

test/testSetup.ts
```diff
@@ -56,8 +57,28 @@ export class TestSetup {
...
     } catch (error) {
-      utilities.log(`Deleting scratch org failed with Error: ${(error as Error).message}`);
+      utilities.log(
+        `Deleting scratch org (or info) failed with Error: ${(error as Error).message}`
+      );
+    }
+  }
+
+  private async deleteScratchOrgInfo(): Promise<void> {
+    if (this.scratchOrgId) {
```
peternhale, line 69: How can we use `scratchOrgAliasName` instead of org Id?

Also, should this function be concerned about an org being reused?

test/testSetup.ts
```diff
@@ -56,8 +57,28 @@ export class TestSetup {
...
+        'data:delete:record',
+        '--sobject',
+        'ScratchOrgInfo',
+        '--where',
+        `Id=${this.scratchOrgId}`
+      );
+      if (sfDataDeleteRecord.exitCode > 0) {
+        const message = `data delete record failed with exit code ${sfDataDeleteRecord.exitCode}\n stderr ${sfDataDeleteRecord.stderr}`;
+        utilities.log(message);
+        throw new Error(message);
```
peternhale, line 80: Why should this cause a test failure?

peternhale (changes requested):
@CristiCanizales I think it would be ok to remove `createInitialProject` now, so if you don't mind including the removal in this PR.

Also, function `createDefaultScratchOrg` could use some attention for code cleanup. Please refactor. Also we are using `orgList` to find the reuse org. Please find a direct way to verify it is present.

test/testSetup.ts
CristiCanizales: `scratchOrgAliasName` is not present in `ScratchOrgInfo`, so we need the ID to identify it

test/testSetup.ts
CristiCanizales: the function is called inside a try block so it wouldn't cause the failure! I can just remove it too 🙂

test/testSetup.ts
CristiCanizales: I could only find the orgId as something to identify the ScratchOrg in ScratchOrgInfo, so I think we can't use `scratchOrgAliasName`. Maybe I'm missing something?

For remote testing I don't think it actually matters as we always run all the tests with `reuseScratchOrg = false`. I was actually wondering if anyone ever sets it to true for local development and if not maybe we can get rid of all that code?

test/testSetup.ts
peternhale: I like removing the try/catch

test/testSetup.ts
CristiCanizales: @peternhale From what I've seen, it's always empty. I just checked again and it's always empty

test/testSetup.ts
CristiCanizales: I'm lost. Removing the try/catch in tearDown that we just added in a recent PR? 🧠

test/testSetup.ts
CristiCanizales: image

peternhale: try `SIGNUPUSERNAME`

test/specs/anInitialSuite.e2e.ts
```diff
@@ -84,7 +84,7 @@ describe('An Initial Suite', async () => {
   });
 
   step('Set up the testing environment', async () => {
-    testSetup = new TestSetup('AnInitialSuite', false);
+    testSetup = new TestSetup('AnInitialSuite');
```
peternhale, line 87:
``` suggestion
    testSetup = new TestSetup('CliCommandTest');
```

peternhale (changes requested):

test/specs/anInitialSuite.e2e.ts
CristiCanizales: This is AnInitialSuite hehe, I changed the name in the clicommands.test.ts

test/testSetup.ts
```diff
@@ -56,8 +57,28 @@ export class TestSetup {
...
-      utilities.log(`Deleting scratch org failed with Error: ${(error as Error).message}`);
+      utilities.log(
+        `Deleting scratch org (or info) failed with Error: ${(error as Error).message}`
+      );
+    }
+  }
+
+  private async deleteScratchOrgInfo(): Promise<void> {
+    if (this.scratchOrgId) {
+      const sfDataDeleteRecord = await utilities.runCliCommand(
```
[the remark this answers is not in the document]
CristiCanizales: Right! Thanks

peternhale (approved):

PR #97, opened 2024-08-05
fix: org browser e2e

peternhale:
@W-14646857@

### DIFF
--- a/test/specs/orgBrowser.e2e.ts
+++ b/test/specs/orgBrowser.e2e.ts
@@ -23,15 +23,23 @@ describe('Org Browser', async () => {
     utilities.log(
       `${testSetup.testSuiteSuffixName} - Check Org Browser is connected to target org`
     );
-    await utilities.executeQuickPick('View: Show Org Browser', utilities.Duration.seconds(5));
+
+    // run org list
+    const orgList = await utilities.orgList();
+    utilities.log(`org list ${orgList.stdout}`);
+    const aliasList = await utilities.aliasList();
+    utilities.log(`alias list ${aliasList.stdout}`);
+    const orgDisplay = await utilities.orgDisplay(testSetup.scratchOrgAliasName!)
+    utilities.log(`org display ${orgDisplay.stdout}`);
+    await utilities.openOrgBrowser();
 
     const orgBrowserLabelEl = await utilities.findElementByText(
       'div',
       'aria-label',
       testSetup.scratchOrgAliasName!
     );
-    utilities.log(`${testSetup.testSuiteSuffixName} - Org Browser is connected to target org`);
     await expect(orgBrowserLabelEl).toBeTruthy();
+    utilities.log(`${testSetup.testSuiteSuffixName} - Org Browser is connected to target org`);
   });
 
   step('Check some metadata types are available', async () => {
@@ -102,7 +110,7 @@ describe('Org Browser', async () => {
     // Verify the deploy was successful
     const successNotificationWasFound = await utilities.notificationIsPresentWithTimeout(
       'SFDX: Deploy This Source to Org successfully ran',
-      utilities.FIVE_MINUTES
+      utilities.Duration.FIVE_MINUTES
     );
     await expect(successNotificationWasFound).toBe(true);
 
@@ -145,7 +153,7 @@ describe('Org Browser', async () => {
 
     const successNotificationWasFound = await utilities.notificationIsPresentWithTimeout(
       'SFDX: Retrieve This Source from Org successfully ran',
-      utilities.FIVE_MINUTES
+      utilities.Duration.FIVE_MINUTES
     );
     await expect(successNotificationWasFound).toBe(true);
   });
@@ -166,7 +174,7 @@ describe('Org Browser', async () => {
     const workbench = await utilities.getWorkbench();
     const successNotificationWasFound = await utilities.notificationIsPresentWithTimeout(
       'SFDX: Retrieve This Source from Org successfully ran',
-      utilities.FIVE_MINUTES
+      utilities.Duration.FIVE_MINUTES
     );
     await expect(successNotificationWasFound).toBe(true);
 
--- a/test/specs/orgBrowser.e2e.ts
+++ b/test/specs/orgBrowser.e2e.ts
@@ -110,7 +110,7 @@ describe('Org Browser', async () => {
   step('Refresh Org Browser and check MyClass is there', async () => {
     utilities.log(`${testSetup.testSuiteSuffixName} - Refresh Apex Classes`);
     // Check MyClass is present under Apex Classes section
-    await utilities.openOrgBrowser(utilities.Duration.seconds(1));
+    await utilities.openOrgBrowser();
     const refreshComponentsButton = await (
       await utilities.findTypeInOrgBrowser('Apex Classes')
     ).$('a[aria-label="SFDX: Refresh Components"]');
--- a/test/specs/orgBrowser.e2e.ts
+++ b/test/specs/orgBrowser.e2e.ts
@@ -26,7 +26,7 @@ describe('Org Browser', async () => {
 
     await utilities.openOrgBrowser();
 
-    await utilities.verifyOrgBrowerIsOpen(
+    await utilities.verifyOrgBrowserIsOpen(
       testSetup.scratchOrgAliasName!
     );
 
--- a/test/specs/orgBrowser.e2e.ts
+++ b/test/specs/orgBrowser.e2e.ts
@@ -24,21 +24,12 @@ describe('Org Browser', async () => {
       `${testSetup.testSuiteSuffixName} - Check Org Browser is connected to target org`
     );
 
-    // run org list
-    const orgList = await utilities.orgList();
-    utilities.log(`org list ${orgList.stdout}`);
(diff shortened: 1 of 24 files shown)


test/utilities/cliCommands.ts
```diff
@@ -108,6 +119,17 @@ export async function orgList(): Promise<SfCommandRunResults> {
   return sfOrgListResult;
 }
 
+export async function aliasList(): Promise<SfCommandRunResults> {
+  const sfAliasListResult = await runCliCommand('alias:list', '--json');
+  if (sfAliasListResult.exitCode > 0) {
+    const message = `org list failed with exit code ${sfAliasListResult.exitCode}\n stderr ${sfAliasListResult.stderr}`;
```
daphne-sfdc, line 125:
```suggestion
    const message = `alias list failed with exit code ${sfAliasListResult.exitCode}\n stderr ${sfAliasListResult.stderr}`;
```

test/utilities/commandPrompt.ts
```diff
@@ -116,9 +117,7 @@ export async function waitForQuickPick(
     {
       timeout: options.timeout?.milliseconds,
       interval: 500, // Check every 500 ms
-      timeoutMsg:
-        options.msg ??
-        `Expected to find option ${pickListItem} before ${options.timeout} milliseconds`
+      timeoutMsg: options.msg ?? `Expected to find option ${pickListItem} before ${options.timeout}`
```
daphne-sfdc, line 120:
```suggestion
      timeoutMsg: options.msg ?? `Expected to find option ${pickListItem} before ${options.timeout} milliseconds`
```

test/utilities/miscellaneous.ts
```diff
@@ -72,6 +74,9 @@ export async function findElementByText(
     timeoutMsg?: string;
   }
 ): Promise<WebdriverIO.Element> {
+  if (!labelText) {
+    throw new Error('labeText must be defined');
```
daphne-sfdc, line 78:
```suggestion
    throw new Error('labelText must be defined');
```

test/utilities/orgBrowser.ts
```diff
@@ -0,0 +1,25 @@
...
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { executeQuickPick } from './commandPrompt.ts';
+import { Duration, findElementByText } from './miscellaneous.ts';
+
+export async function openOrgBrowser(wait: Duration = Duration.seconds(1)): Promise<void> {
+  await executeQuickPick('View: Show Org Browser', wait);
+}
+
+export async function verifyOrgBrowerIsOpen(label: string): Promise<void> {
```
daphne-sfdc, line 14: Fix all references to this function too.

```suggestion
export async function verifyOrgBrowserIsOpen(label: string): Promise<void> {
```

test/utilities/workbench.ts
```diff
@@ -55,6 +55,13 @@ export async function zoomReset(wait: Duration = Duration.seconds(1)): Promise<v
   await executeQuickPick('View: Reset Zoom', wait);
 }
 
+export async function openNewTerminal(checkIfOpen = false): Promise<void> {
+  if (checkIfOpen) {
+    // TODO: do something here???
+  }
```
daphne-sfdc, line 61: I assume you have the if statement because you want to reuse the existing Terminal if one already exists?  In that case try `Output: Focus on Terminal View` for the TODO.  (By the way, `Output: Focus on Terminal View` does create a new Terminal if one does not already exist, so the if statement could be removed if this is indeed the intended use case.)

But if we do want to reuse Terminal windows, then the function name `openNewTerminal` sounds misleading, as it sounds like a new Terminal window should be created regardless if one already exists.

test/specs/orgBrowser.e2e.ts
```diff
@@ -113,60 +110,58 @@ describe('Org Browser', async () => {
   step('Refresh Org Browser and check MyClass is there', async () => {
     utilities.log(`${testSetup.testSuiteSuffixName} - Refresh Apex Classes`);
     // Check MyClass is present under Apex Classes section
-    await utilities.executeQuickPick('View: Show Org Browser', utilities.Duration.seconds(5));
+    await utilities.openOrgBrowser(utilities.Duration.seconds(1));
```
daphne-sfdc, line 113: Default is already 1 second.

```suggestion
    await utilities.openOrgBrowser();
```

test/specs/orgBrowser.e2e.ts
```diff
@@ -113,60 +110,58 @@ describe('Org Browser', async () => {
...
-      'li',
-      'title',
-      'Retrieve This Source from Org'
+    const retrieveSourceButton = await myClassLabelEl.$(
+      '//a[@aria-label="Retrieve This Source from Org"]'
     );
-    utilities.debug(`button 2 ${JSON.stringify(retrieveSourceButton)}`);
-    await retrieveSourceButton.click();
+
+    await browser.elementHover(retrieveSourceButton.elementId);
```
daphne-sfdc, line 139: FYI that `elementHover` produces this error when I run the E2E test locally:
```
[chrome 124.0.6367.243 mac #0-0] 1) Org Browser Retrieve This Source from Org
[chrome 124.0.6367.243 mac #0-0] unknown command: unknown command: session/2271a149135e0a50b27449d6317a29a7/element/f.EEE8F9C61C76DF8548C27053A71E75AC.d.CD3FE8E0990D4FB67C7FAA8F1DCD5203.e.859/hover
[chrome 124.0.6367.243 mac #0-0] unknown command: unknown command: unknown command: session/2271a149135e0a50b27449d6317a29a7/element/f.EEE8F9C61C76DF8548C27053A71E75AC.d.CD3FE8E0990D4FB67C7FAA8F1DCD5203.e.859/hover
[chrome 124.0.6367.243 mac #0-0]     at async Context.<anonymous> (file:///Users/daphne.yang/Development/salesforcedx-vscode-automation-tests/test/specs/orgBrowser.e2e.ts:139:5)
```

So I think this function exists in WDIO but is not actually supported for local runs.  (I haven't tried on GHA.)

test/utilities/workbench.ts
peternhale: Its a TODO for now.

test/testSetup.ts
```diff
@@ -105,14 +105,18 @@ export class TestSetup {
...
+      ? Env.getInstance().useExistingProject
+      : path.join(this.tempFolderPath, this.tempProjectName);
     utilities.log(
       `${this.testSuiteSuffixName} - creating project files in ${this.projectFolderPath}`
     );
 
-    // Remove the project folder, just in case there are stale files there.
-    if (fs.existsSync(this.projectFolderPath)) {
-      utilities.removeFolder(this.projectFolderPath);
+    // Remove the project folder, just in case there are stale files there, but only if it is not an existing projet.
```
daphne-sfdc, line 115:
```suggestion
    // Remove the project folder, just in case there are stale files there, but only if it is not an existing project.
```

test/utilities/commandPrompt.ts
```diff
@@ -184,3 +183,8 @@ export async function clickFilePathOkButton(): Promise<void> {
   }
   await pause(Duration.seconds(2));
 }
+
+export async function setDefaultOrg(targetOrg: string): Promise<void> {
+  const inputBox = await executeQuickPick('SFDX: Set a Default Org');
+  await findQuickPickItem(inputBox, targetOrg, false, true);
+}
```
daphne-sfdc, line 190: I don't think this helper function should belong in `commandPrompt.ts`.  Can we make a new utility file `authentication.ts` or at least move the function to `miscellaneous.ts`?

test/testSetup.ts
```diff
@@ -224,103 +232,84 @@ export class TestSetup {
...
 
-    utilities.log(`${this.testSuiteSuffixName} - constructing scratchOrgAliasName...`);
+    utilities.debug(`${this.testSuiteSuffixName} - constructing scratchOrgAliasName...`);
     // Org alias format: TempScratchOrg_yyyy_mm_dd_username_ticks_testSuiteSuffixName
     const currentDate = new Date();
-    const ticks = currentDate.getTime();
-    const day = ('0' + currentDate.getDate()).slice(-2);
-    const month = ('0' + (currentDate.getMonth() + 1)).slice(-2);
+    const day = currentDate.getDate().toString().padStart(2, '0');
+    const month = (currentDate.getMonth() + 1).toString().padStart(2, '0');
```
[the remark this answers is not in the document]
peternhale: Function `currentDate.getMonth()` returns a 0 based index, so adding 1 returns it to reality.

test/specs/orgBrowser.e2e.ts
peternhale: What do you mean "local runs?" Its my understanding that wdio works the same wherever it is run.

daphne-sfdc: By "local runs" I meant running the E2E test locally on my computer.  In that case, this function should be broken on a GHA run too.

daphne-sfdc (approved):
Approved! 🎉 

Passing E2E test run for org browser here: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/10291823010 ✅

PR #100, opened 2024-08-20
feat: test using an existing project

mingxuanzhangsfdx:


test/specs/exampleUsingProject.e2e.ts
```diff
@@ -0,0 +1,15 @@
...
+import * as utilities from '../utilities/index.ts';
+
+describe('an example using existing project', async () => {
+  let testSetup: TestSetup;
+
+  step('Set up the testing environment', async () => {
+    testSetup = new TestSetup('usingExistingProject');
+    await testSetup.setUp('developer', utilities.repoKeywords.DeployInv);
+    await utilities.reloadAndEnableExtensions();
+    await expect(1).toBe(false); // make it false to see screenshots
```
mingxuanzhangsfdx, line 12: I made the tests fail deliberately since I wanted to see if the project is really opened up.

test/utilities/constants.ts
```diff
@@ -17,3 +17,13 @@ export const WORKSPACE_SETTING_KEYS = {
...
 
 export type LogLevel = (typeof LOG_LEVELS)[number];
+
+// the name of repos
+export enum repoKeywords {
+  DeployInv = 'DeployInv'
+}
+
+// a map from a repo name to git url
+export const projectMaps = new Map<repoKeywords, string>([
```
mingxuanzhangsfdx, line 27: So we are going to maintain such a map to make e2e tests download them from Github.

test/testSetup.ts
```diff
@@ -11,34 +11,37 @@ import { DefaultTreeItem, InputBox, QuickOpenBox } from 'wdio-vscode-service';
...
-  public tempFolderPath: string | undefined = undefined;
+  public tempFolderPath: string;
   public projectFolderPath: string | undefined;
   private prompt: QuickOpenBox | InputBox | undefined;
   public scratchOrgAliasName: string | undefined;
   public scratchOrgId: string | undefined;
 
   public constructor(testSuiteSuffixName: string) {
     this.testSuiteSuffixName = testSuiteSuffixName;
+    this.tempFolderPath = path.join(__dirname, '..', 'e2e-temp');
```
mingxuanzhangsfdx, line 30: tempFolderPath should be initialized in constructor. The previous impl made code more complicated.

test/specs/exampleUsingProject.e2e.ts
```diff
@@ -0,0 +1,15 @@
+import { step } from 'mocha-steps';
+import { TestSetup } from '../testSetup.ts';
+import * as utilities from '../utilities/index.ts';
+
+describe('an example using existing project', async () => {
+  let testSetup: TestSetup;
+
+  step('Set up the testing environment', async () => {
+    testSetup = new TestSetup('usingExistingProject');
+    await testSetup.setUp('developer', utilities.repoKeywords.DeployInv);
```
mingxuanzhangsfdx, line 10: You need to set up the test by adding the repo name. After this, you will have the project you want open in VSCode

test/specs/exampleUsingProject.e2e.ts
mingxuanzhangsfdx: And the screenshots [here](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/10466350108) tell.

test/specs/exampleUsingProject.e2e.ts
mingxuanzhangsfdx: if you do not need to use scratch orgs, you can still set up the test by using setUpTestingEnvironment and createOrUseProject to initialize your test

test/testSetup.ts
```diff
@@ -117,21 +125,20 @@ export class TestSetup {
       if (fs.existsSync(this.projectFolderPath)) {
         utilities.removeFolder(this.projectFolderPath);
       }
-    }
-
-    // Now create the temp folder.  It should exist but create the folder if it is missing.
-    if (!fs.existsSync(this.tempFolderPath)) {
-      utilities.createFolder(this.tempFolderPath);
+      // Now create the temp folder.  It should exist but create the folder if it is missing.
+      if (!fs.existsSync(this.tempFolderPath)) {
```
mingxuanzhangsfdx, line 129: the temp folder is needed only when you do not use an existing project/

test/utilities/constants.ts
```diff
@@ -17,3 +17,13 @@ export const WORKSPACE_SETTING_KEYS = {
 export const LOG_LEVELS = ['trace', 'debug', 'info', 'warn', 'error', 'silent'] as const;
 
 export type LogLevel = (typeof LOG_LEVELS)[number];
+
+// the name of repos
+export enum repoKeywords {
+  DeployInv = 'DeployInv'
+}
```
peternhale, line 24: Changing the enum in this fashion eliminates the need for the map
``` suggestion
// the name of repos
export enum repoKeywords {
  DeployInv = 'https://github.com/mingxuanzhangsfdx/DeployInv'
}
```

test/utilities/constants.ts
peternhale: We have been discussing the need to expose our testing framework  How do you see us supporting such a structure when customers are using our framework?

The current structure of test/workspace initialization is configuration based, while this solution makes it a code based.

I am in favor of tooling that simplifies the test authors productivity, so I look forward to alternative suggestions. 

Please keep in mind local development and CI/CD.

test/testSetup.ts
```diff
@@ -99,11 +102,16 @@ export class TestSetup {
...
-  public async setUpTestingEnvironment(): Promise<void> {
+  public async setUpTestingEnvironment(repo?: utilities.repoKeywords): Promise<void> {
     utilities.log('');
     utilities.log(`${this.testSuiteSuffixName} - Starting setUpTestingEnvironment()...`);
 
-    this.tempFolderPath = path.join(__dirname, '..', 'e2e-temp');
+    if (repo) {
+      const repoPath = path.join(__dirname, '..', repo);
+      await gitClone(utilities.projectMaps.get(repo) as string, repoPath); // clone the project
+      Env.getInstance().useExistingProject = repoPath; // set the existing project file path
```
peternhale, line 112: Not sure changing the value inside env settings is the right thing to do.

We run our tests one at a time, so there are no side effects here, but thinking into the future when customers might be using this test framework where they might be running a number of tests at a time, modifying the env setting might a negative effect on a test expecting a separate existing project.

test/utilities/constants.ts
mingxuanzhangsfdx: we still need a string to construct the path

test/specs/constants.ts
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright (c) 2023, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ * We store the salesforcedx-vscode specific constants here. 
+ */
+
+// the name of repos
+export enum repoKeywords {
```
mingxuanzhangsfdx, line 10: I still want to define the enum like this, since we are going to need the name for folder path.

test/testSetup.ts
```diff
@@ -99,11 +103,16 @@ export class TestSetup {
...
-  public async setUpTestingEnvironment(): Promise<void> {
+  public async setUpTestingEnvironment(repo?: constants.repoKeywords): Promise<void> {
     utilities.log('');
     utilities.log(`${this.testSuiteSuffixName} - Starting setUpTestingEnvironment()...`);
 
-    this.tempFolderPath = path.join(__dirname, '..', 'e2e-temp');
+    if (repo) {
+      const repoPath = path.join(__dirname, '..', repo);
+      await gitClone(constants.projectMaps.get(repo)!, repoPath); // clone the project
+      Env.getInstance().useExistingProject = repoPath; // set the existing project file path, value will be overriden even if USE_EXISTING_PROJECT_PATH exists
```
peternhale, line 113: Modifying this var will have a negative effect on other tests that might run along with the one that is using a specific repository.

The value set wuen initilizing env var should not be modified to suit the needs of this change.

While establishing the project that will be used
- The test author wants a new project
- The test author wants a project cloned into the test
- The test author will use the project specified by Env.getInstance().useExistingProject
- The configuration of one test shall not affect the state of other tests

test/testSetup.ts
```diff
@@ -99,11 +103,16 @@ export class TestSetup {
...
   }
 
-  public async setUpTestingEnvironment(): Promise<void> {
+  public async setUpTestingEnvironment(repo?: constants.repoKeywords): Promise<void> {
     utilities.log('');
     utilities.log(`${this.testSuiteSuffixName} - Starting setUpTestingEnvironment()...`);
 
-    this.tempFolderPath = path.join(__dirname, '..', 'e2e-temp');
+    if (repo) {
+      const repoPath = path.join(__dirname, '..', repo);
```
peternhale, line 111: Why not use the `this.tempFolderPath` as the root dir for the clone?
```suggestion
      const repoPath = path.join(this.tempFolderPath, repo);
```

test/testSetup.ts
```diff
@@ -99,11 +103,16 @@ export class TestSetup {
...
+      await gitClone(constants.projectMaps.get(repo)!, repoPath); // clone the project
+      Env.getInstance().useExistingProject = repoPath; // set the existing project file path, value will be overriden even if USE_EXISTING_PROJECT_PATH exists
+      utilities.log(`Env.getInstance().useExistingProject - ${Env.getInstance().useExistingProject}`)
+    }
 
     this.projectFolderPath = Env.getInstance().useExistingProject
```
peternhale, line 117: Should we support the possibility that the env var could a repo URL or a directory?

peternhale: One other are that needs to be rethought is the code that cleans up a new project if present and teardown.

Let me knw your thoughts on this.

PR #44, opened 2023-08-17
debugApexTests.test.ts runs successfully on remote

daphne-sfdc:
### How to test:
(1) Go to the [End to End Tests Workflow](https://github.com/forcedotcom/salesforcedx-vscode/actions/workflows/e2e.yml) in salesforcedx-vscode repo
(2) Paste the PR's branch (daphne/debugApexTests-on-remote) in the second field of the Run workflow dropdown (Set the branch to use for automation tests)
(3) Choose debugApexTests.e2e.ts test from the dropdown to select the test to run.

AC:
Test suite should pass.

### Key changes made:

(1) Added logic to trigger refreshment of the APEX TESTS section of the sidebar if tests do not show up.

(2) Used notificationIsPresentWithTimeout() instead of notificationIsPresent().  As a result of the change, removed all calls to waitForNotificationToGoAway().

(3) Uncovered the 'Debug' button that was previously covered by the command palette due to small VSCode window size in Remote.

(4) Used the 'Testing: Focus on Apex Tests View' prompt in the command palette to open the Testing view instead of using the sidebar.

### Successful runs:
• [338](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/5895691735)
• [339](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/5895695565)
• [340](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/5895698456)

@W-13743990@

PR #103, opened 2024-09-05
Feat: refactor testSetup class with project shape config

mingxuanzhangsfdx:
1. Refactor the testSetup class with a config called TestReqConfig that provides options of
    1. project shape (new, none, named and any)
    2. if an authorized org is required
    3. scratch org edition (developer or enterprise)
    4. the extensions to be installed and checked
    5. test suffix name

2. Refactor all tests to consume the new framework

3. Add new tests for 
    1. SFDX: create project command (example of none project shape) can be run on branch `mz/mdDeployRetrieve` in vsce as part of miscellaneous
    2. Deploy and Retrieve MD file (example of named project shape) can be run on branch `mz/mdDeployRetrieve` in vsce as part of deploy and retrieve test


peternhale (changes requested):

test/specs/miscellaneous.e2e.ts
```diff
@@ -44,7 +50,7 @@ describe('Miscellaneous', async () => {
     const workbench = await utilities.getWorkbench();
     const commandName =
       EnvironmentSettings.getInstance().vscodeVersion === 'stable' ||
-      semver.gte(EnvironmentSettings.getInstance().vscodeVersion, '1.92.0')
+        semver.gte(EnvironmentSettings.getInstance().vscodeVersion, '1.92.0')
```
mingxuanzhangsfdx, line 53: The logic here failed locally when default _vscodeVersion is still stable and the running version of VSCode is 1.91.0. @daphne-sfdc

test/RefactoredTestSetup.ts
```diff
@@ -0,0 +1,360 @@
...
+            const localProjName = utilities.getFolderName(projectConfig.folderPath);
+            if (localProjName != repoName) {
+              this.throwError(`the local project ${projectConfig.githubRepoUrl} does not match the required Github repo ${projectConfig.githubRepoUrl}`);
+            } else {
+              // if it is a match, use the local folder directly. Local dev use only.
+              this.projectFolderPath = projectConfig.folderPath;
+            }
+          } else {
+            // clone the project from Github URL directly
+            this.projectFolderPath = path.join(this.tempFolderPath, repoName!);
```
CristiCanizales, line 97: repoName could be null if something fails with `utilities.getRepoNameFromUrl()`

CristiCanizales (changes requested):
Great work Mingxuan! I really like the new RefactoredTestSetup and how we can do or don't do stuff depending on each test suite needs :awesome:

I just have some comments and a couple requests

test/specs/miscellaneous.e2e.ts
mingxuanzhangsfdx: Hmmm It works on my computer though. Nvm

test/RefactoredTestSetup.ts
mingxuanzhangsfdx: it is checked in `gitRepoExists`, which will see if it is a valid github url or not. If it is, then repoName must exist. Does this make sense?

daphne-sfdc (changes requested):

test/utilities/extensionUtils.ts
```diff
@@ -425,3 +426,31 @@ export async function findExtensionsInRunningExtensionsList(
   // limit runningExtensions to those whose property extensionId is in the list of extensionIds
   return runningExtensions.filter((extension) => extensionIds.includes(extension.extensionId));
 }
+
+export async function verifyProjectCreated(projectName: string) {
```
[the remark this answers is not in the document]
daphne-sfdc: I like `sideBar.ts` because the verification all happens in the sidebar.

test/specs/anInitialSuite.e2e.ts
```diff
@@ -83,16 +83,19 @@ describe('An Initial Suite', async () => {
...
+    isOrgRequired: false,
+    testSuiteSuffixName: 'AnInitialSuite'
+  }
+
+  step('Set up the testing environment and verify our extensions are loaded after creating an SFDX project', async () => {
+    await testSetup.setUp(testReqConfig);
   });
 
-  step('Verify our extensions are loaded after creating an SFDX project', async () => {
+  step('', async () => {
```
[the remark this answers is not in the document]
daphne-sfdc: And the previous step would just be "Set up the testing environment", like it was before.

test/RefactoredTestSetup.ts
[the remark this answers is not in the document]
mingxuanzhangsfdx: I changed the class name following camel naming but due to the limitation of git, renaming for the file cannot be tracked. But I think it is not a big deal since we will finally replace testSetup with it.

test/RefactoredTestSetup.ts
```diff
@@ -0,0 +1,360 @@
...
+  }
+
+  public async tearDown(): Promise<void> {
+    await this.checkForUncaughtErrors();
+    try {
+      await utilities.deleteScratchOrg(this.scratchOrgAliasName);
+      await this.deleteScratchOrgInfo();
+      this.projectFolderPath = undefined;
+      this.scratchOrgAliasName = undefined;
+      this.scratchOrgId = undefined;
```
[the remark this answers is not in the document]
mingxuanzhangsfdx: Each test will maintain its own refactoredTestSetup class. Good catch, I think they are useless.

test/RefactoredTestSetup.ts
```diff
@@ -0,0 +1,360 @@
...
+        if (projectConfig.githubRepoUrl) {
+          // verify if folder matches the github repo url
+          const repoExists = await utilities.gitRepoExists(projectConfig.githubRepoUrl);
+          if (!repoExists) {
+            this.throwError(`Repository does not exist or is inaccessible: ${projectConfig.githubRepoUrl}`);
+          }
+          const repoName = utilities.getRepoNameFromUrl(projectConfig.githubRepoUrl);
+          projectName = repoName;
+          if (projectConfig.folderPath) {
+            const localProjName = utilities.getFolderName(projectConfig.folderPath);
```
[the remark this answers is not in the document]
mingxuanzhangsfdx: Yes. If both local project and github url are defined, we will see if local project refers to the same project on github. If so, we are use local project directly and do not need to clone from github.

test/RefactoredTestSetup.ts
```diff
@@ -0,0 +1,360 @@
...
+          // missing info, throw an error
+          this.throwError(`githubRepoUrl is required for named project shape`);
+        }
+        break;
+
+      case ProjectShapeOption.ANY:
+        // ANY: workspace is designated to open when wdio is initialized
+        if (projectConfig.folderPath) {
+          this.projectFolderPath = projectConfig.folderPath;
+          projectName = utilities.getFolderName(projectConfig.folderPath);
```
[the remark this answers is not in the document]
mingxuanzhangsfdx: 1. On GHA, we can always do git clone before wdio is started
2. On local, we can also always clone before we run test locally

We can support do git clone when wdio is run but I do not see a necessity for it, and also when running on wdio, we try to avoid dummy workspace opening. So I prefer to only support starting with the target project once wdio is run.

test/RefactoredTestSetup.ts
```diff
@@ -0,0 +1,360 @@
...
+    utilities.log(`Project folder to open: ${this.projectFolderPath}`);
+    await utilities.openFolder(this.projectFolderPath!);
+    // Verify the project was created and was loaded.
+    await utilities.verifyProjectCreated(projectName ?? this.tempProjectName);
+  }
+
+  throwError(message: string) {
+    utilities.log(message);
+    throw new Error(message);
+  }
```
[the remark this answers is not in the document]
mingxuanzhangsfdx: Yeah it should be. I thought if I do not explicitly define the access, it was private by default. But I am wrong it is public by default.

test/specs/apexReplayDebugger.e2e.ts
```diff
@@ -7,20 +7,26 @@
...
 const CMD_KEY = process.platform === 'darwin' ? Key.Command : Key.Control;
 
 describe('Apex Replay Debugger', async () => {
   let prompt: QuickOpenBox | InputBox;
-  let testSetup: TestSetup;
+  const testSetup = new RefactoredTestSetup();
+  const testReqConfig: utilities.TestReqConfig = {
+    projectConfig: {
+      projectShape: utilities.ProjectShapeOption.NEW,
+    },
```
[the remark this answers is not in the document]
mingxuanzhangsfdx: Sure! We definitely can optimize the test for ard-specific test to only focus on it. And what's more, we might even consider to use a normal dev org with apex classes deployed already. So we do not need to create a scratch org and push them again and again.

test/utilities/extensionUtils.ts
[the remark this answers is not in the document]
mingxuanzhangsfdx: Thanks for the catch!

test/utilities/types.ts
```diff
@@ -5,10 +5,44 @@
...
+  NAMED // tests will be run on a well-know test project once wdio is initialized
+}
+
+export type ProjectConfig = {
+  /* The shape of the test project that the test runs on*/
+  projectShape: ProjectShapeOption,
+  /* The local path to the project if the project shape is any or named */
+  folderPath?: string,
+  /* The url of the github repo, only exists when project shape is named*/
+  githubRepoUrl?: string,
```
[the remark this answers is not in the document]
mingxuanzhangsfdx: Like I mentioned above, we do not need it if it is "any". Since the project should have been always existed locally.

test/utilities/types.ts
```diff
@@ -5,10 +5,44 @@
...
+
+export type TestReqConfig = {
+  /* The project shape and the local path to the project */
+  projectConfig: ProjectConfig,
+  /* If org is required for the test. If not, do not need to create and log into the scratch org */
+  isOrgRequired: boolean, //
+  /* The edition of the scratch org to be created, only specified when isOrgRequired is true*/
+  scratchOrgEdition?: OrgEdition,
+  /* The required extensions */
+  excludedExtensions?: ExtensionId[],
```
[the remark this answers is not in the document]
mingxuanzhangsfdx: Thanks for catching this! I was wrong about the explanation - since it will be used in the function `utilities.installExtensions` it should be extensions that do not need to be installed. Updated!

test/RefactoredTestSetup.ts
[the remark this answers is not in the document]
daphne-sfdc: I meant, what if you take the project url as an additional input in the workflow dispatch for the "any" case?  The project from that input url would be cloned before WDIO is started.

test/utilities/apexUtils.ts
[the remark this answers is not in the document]
mingxuanzhangsfdx: It is a correct direction but I would like to put it into another story. The two functions both have several side effects (output panel verification, setting breakpoints, specify text to populate etc...). And that comes to refactor all functions in apexUtils. Does that sound good?

test/RefactoredTestSetup.ts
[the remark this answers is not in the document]
mingxuanzhangsfdx: Do you mean something I did [here](https://github.com/forcedotcom/salesforcedx-vscode/pull/5730/files#diff-9ae54a92531817cc2244971629d759a5ceb94fb6992b665fcfd57fdbc46af0ad)?

test/utilities/apexUtils.ts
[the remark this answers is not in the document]
CristiCanizales: I didn't mean to directly use that function but to do what we do there, so instead of `await browser.keys(['Enter']);` do `await inputBox.confirm();`

test/RefactoredTestSetup.ts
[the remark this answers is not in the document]
daphne-sfdc: Yes! Use the `githubRepoUrl` input as the “any” repo that you want to clone for the current E2E test. 🙂

test/refactoredTestSetup.ts
```diff
@@ -0,0 +1,336 @@
...
+
+export class refactoredTestSetup {
+  public testSuiteSuffixName: string = '';
+  public tempFolderPath = path.join(__dirname, '..', 'e2e-temp');
+  public projectFolderPath: string | undefined;
+  private aliasAndUserNameWereVerified = false;
+  public scratchOrgAliasName: string | undefined;
+  public scratchOrgId: string | undefined;
+
+  public constructor() {
```
peternhale, line 25: What do you think about passing test config in the constructor and then referencing it in the other functions? 

My thinking would be to limit the possibility of changing the test config using instance methods.

With the current impl I can do things like
```
const ts = new refactoredTestSetup();
await ts.setup(config1);
await ts.setup(config2);
await ts.setUpTestingWorkspace(projectConfig1);
await ts.setUpTestingWorkspace(projectConfig2);
```
While nonsensical, it does show that the test setup is not controlling the config of the test.

This a balance between flexibility and config control, so think about finding a balance.

This also could be done with the factory pattern.

```
public class TestSetup {
  private constructor(public readonly testReqConfig: utilities.TestReqConfig) {}
  public static async setup(testReqConfig: utilities.TestReqConfig): Promise<void> {
    // do setup stuff
  }
  // instance methods that we allow on the test setup
  public async teardown() {}
}
```
then
```
const ts = await TestSetup.setup(testConfig);
// do test related stuff and mutating the test project context is not allowed via `ts`
```

Our discussions thus far have been around one test/one project. How do we and why would we handle changing the workspace to an entirely different project?

test/refactoredTestSetup.ts
```diff
@@ -0,0 +1,336 @@
...
+          this.throwError(`githubRepoUrl is required for named project shape`);
+        }
+        break;
+
+      case ProjectShapeOption.ANY:
+        // ANY: workspace is designated to open when wdio is initialized
+        if (projectConfig.folderPath) {
+          this.projectFolderPath = projectConfig.folderPath;
+          projectName = utilities.getFolderName(projectConfig.folderPath);
+        } else this.throwError('folder path is required for named project!');
```
peternhale, line 108: @mingxuanzhangsfdx error message text is incorrect.

test/refactoredTestSetup.ts
```diff
@@ -0,0 +1,336 @@
...
+          }
+        } else {
+          // missing info, throw an error
+          this.throwError(`githubRepoUrl is required for named project shape`);
+        }
+        break;
+
+      case ProjectShapeOption.ANY:
+        // ANY: workspace is designated to open when wdio is initialized
+        if (projectConfig.folderPath) {
```
peternhale, line 105: I thought the fallback for any if no project folder is provided would be to create a `new` project, instead of failing.

test/refactoredTestSetup.ts
```diff
@@ -0,0 +1,336 @@
...
+            const localProjName = utilities.getFolderName(projectConfig.folderPath);
+            if (localProjName != repoName) {
+              this.throwError(`the local project ${projectConfig.githubRepoUrl} does not match the required Github repo ${projectConfig.githubRepoUrl}`);
+            } else {
+              // if it is a match, use the local folder directly. Local dev use only.
+              this.projectFolderPath = projectConfig.folderPath;
+            }
+          } else {
+            // clone the project from Github URL directly
+            this.projectFolderPath = path.join(this.tempFolderPath, repoName!);
```
peternhale, line 94: use of `repoName!` is discouraged.

test/refactoredTestSetup.ts
```diff
@@ -0,0 +1,336 @@
...
+      case ProjectShapeOption.ANY:
+        // ANY: workspace is designated to open when wdio is initialized
+        if (projectConfig.folderPath) {
+          this.projectFolderPath = projectConfig.folderPath;
+          projectName = utilities.getFolderName(projectConfig.folderPath);
+        } else this.throwError('folder path is required for named project!');
+        return;
+
+      case ProjectShapeOption.NONE:
+        // NONE: blank project by default
```
peternhale, line 112: Not a blank project, but no project at all.

test/refactoredTestSetup.ts
```diff
@@ -0,0 +1,336 @@
...
+        /* create the e2e-temp folder to benefit further testing */
+        if (!fs.existsSync(this.tempFolderPath)) {
+          utilities.createFolder(this.tempFolderPath);
+        }
+        this.projectFolderPath = path.join(this.tempFolderPath, this.tempProjectName);
```
peternhale, line 117: What is the value of `this.tempProjectName` at this moment? Seems like it would be empty string. Is this what we want. Also, if the test never creates a project, where will screnshots of failures be saved and associated with the current test.

test/utilities/apexUtils.ts
```diff
@@ -18,26 +19,24 @@ export async function createApexClass(
...
-  await browser.keys(['Enter']);
+  await inputBox.confirm();
   await pause(Duration.seconds(1));
 
   // Select the default directory (press Enter/Return).
-  await browser.keys(['Enter']);
-  await pause(Duration.seconds(1));
-
+  await inputBox.confirm();
+  await pause(Duration.seconds(2)); // have to hard pause for text editor to be ready
```
peternhale, line 31: What else can be used to wait for the new file to be ready for use?

test/utilities/apexUtils.ts
```diff
@@ -18,26 +19,24 @@ export async function createApexClass(
...
   // Modify class content
   const textEditor = await getTextEditor(workbench, name + '.cls');
   await textEditor.setText(classText);
   await textEditor.save();
   if (breakpoint) {
     await textEditor.toggleBreakpoint(breakpoint);
   }
-  await pause(Duration.seconds(1));
+  await pause(Duration.seconds(2));
```
peternhale, line 39: What is the pause for?

test/refactoredTestSetup.ts
```diff
@@ -0,0 +1,336 @@
+/*
```
peternhale, line 1: @mingxuanzhangsfdx Would you mind taking the accumulated changes to this file and apply them to TestSetup and then remove the new file. Tracking changes across two different files is a pain.

test/utilities/types.ts
```diff
@@ -5,10 +5,44 @@
...
+
+export type TestReqConfig = {
+  /* The project shape and the local path to the project */
+  projectConfig: ProjectConfig,
+  /* If org is required for the test. If not, do not need to create and log into the scratch org */
+  isOrgRequired: boolean, //
+  /* The edition of the scratch org to be created, only specified when isOrgRequired is true*/
+  scratchOrgEdition?: OrgEdition,
+  /* The extensions that do not need to be installed */
+  excludedExtensions?: ExtensionId[],
```
peternhale, line 44: I know I wrote this initially excluding extensions, but just does not make sense in the larger context of a test framework. This would not need to be done in your current work, because "how to configure extensions?" deserves discussion.

test/environmentSettings.ts
```diff
@@ -71,6 +76,8 @@ export class EnvironmentSettings {
       : this._logLevel;
     this._javaHome = process.env.JAVA_HOME || this._javaHome;
     this.useExistingProject = process.env.USE_EXISTING_PROJECT_PATH;
```
peternhale, line 78: @mingxuanzhangsfdx did you want to get rid of `USE_EXISTING_PROJECT_PATH` in favor of `LOCAL_PROJECT_PATH`. If not can you justify having both?

peternhale (changes requested):

test/refactoredTestSetup.ts
```diff
@@ -0,0 +1,336 @@
...
+    await utilities.installExtensions(testReqConfig.excludedExtensions);
+    await utilities.reloadAndEnableExtensions();
+    /* The expected workspace will be open up after setUpTestingWorkspace */
+    await this.setUpTestingWorkspace(testReqConfig.projectConfig);
+    if (testReqConfig.projectConfig.projectShape != ProjectShapeOption.NONE) {
+      await utilities.verifyExtensionsAreRunning(utilities.getExtensionsToVerifyActive());
+      const scratchOrgEdition = testReqConfig.scratchOrgEdition || 'developer';
+      this.updateScratchOrgDefWithEdition(scratchOrgEdition);
+      if (process.platform === 'darwin') this.setJavaHomeConfigEntry();
+      // Extra config needed for Apex LSP on GHA
```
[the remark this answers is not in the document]
mingxuanzhangsfdx: it should be together with line #44, updated.

test/refactoredTestSetup.ts
mingxuanzhangsfdx: You are right the projectFolderPath here should be undefined. 

The screenshots should be saved according to the test title and the test step name. So it has nothing to do with if a project is open or not.

test/utilities/apexUtils.ts
mingxuanzhangsfdx: The function `createApexClass` can be called multiple times in a row in its consumers such as `createApexClassWithTest` and `createApexClassWithBugs`. The pause here is to give enough time between saving the text and calling the next `SFDX: Create Apex Class`. I noticed that pauses with short time may cause failure in popping up the command palette. Increasing the pause time could help reduce flappers like that.

test/utilities/apexUtils.ts
mingxuanzhangsfdx: Great idea - I used `browswer.waitUntil`. Check out the function `checkFileOpen` I just implemented.

test/refactoredTestSetup.ts
mingxuanzhangsfdx: When I started the effort, I wanted to make sure my change did not affect the tests that I have not taken care of because there are a lot tests left. Sorry for the confusion.

test/utilities/types.ts
mingxuanzhangsfdx: Yes I also have some ideas, such as 
1. Specifying the extensions we want to include in, not the ones we want to exclude, 
2. Making core extension as a mandatory one, and others as optional.

Since we actually have never utilized such function ever, we can discuss this later.

test/environmentSettings.ts
mingxuanzhangsfdx: Yes I do. I want to replace with `LOCAL_PROJECT_PATH` for better clarification. Just updated

test/utilities/lwcUtils.ts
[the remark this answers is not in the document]
mingxuanzhangsfdx: updated

test/refactoredTestSetup.ts
mingxuanzhangsfdx: An example of calling `setUp` multiple times within one test is `pushAndPull.e2e.ts` where we create a second project to connect to the same scratch org and pull the source. So with this case as an example, I think testSetup should be associated with one test rather than one project setting.

The reason why I do not use constructor is mostly that we cannot call async functions within it. 

Does this make sense to you?

test/utilities/fileSystem.ts
```diff
@@ -64,3 +64,24 @@ export function getVsixFilesFromDir(vsixDir: string): string[] {
     path.join(vsixDir, vsixFile)
   );
 }
+
+/**
+ * Check if the given path refers to a directory
```
peternhale, line 69:
```suggestion
 * Return folder name if given path is a directory, otherwise return null
```

test/environmentSettings.ts
```diff
@@ -120,21 +125,34 @@ export class EnvironmentSettings {
...
       return;
     }
     if (!fs.existsSync(projectPath)) {
       throw new Error(`Project path for "${projectPath}" does not exist`);
     }
+    this._localProjectPath = projectPath;
+  }
+
+  public set githubProjectUrl(githubUrl: string | undefined) {
+    const projectPath = githubUrl ?? process.env.GITHUB_PROJECT_URL;
```
peternhale, line 149: Would you mind changing the local var name projectPath to something more suitable for handling URL?

peternhale (changes requested):
@mingxuanzhangsfdx looks much better, thank you. I did request more changes though. Let me know what you think.

test/testSetup.ts
```diff
@@ -1,55 +1,54 @@
...
+    await utilities.installExtensions(testReqConfig.excludedExtensions);
     await utilities.reloadAndEnableExtensions();
-    await utilities.verifyExtensionsAreRunning(utilities.getExtensionsToVerifyActive());
-    await this.authorizeDevHub();
-    await this.createDefaultScratchOrg();
-    utilities.log(`${this.testSuiteSuffixName} - ...finished TestSetup.setUp()`);
-    utilities.log('');
+    /* The expected workspace will be open up after setUpTestingWorkspace */
+    await this.setUpTestingWorkspace(testReqConfig.projectConfig);
+    if (testReqConfig.projectConfig.projectShape !== ProjectShapeOption.NONE) {
```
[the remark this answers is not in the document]
mingxuanzhangsfdx: The function is to make sure the expected workspace is open after being executed. The below if block is to set up scratch org, Java. So I would prefer separating them.

test/testSetup.ts
```diff
@@ -60,114 +59,101 @@ export class TestSetup {
...
-      }
+      default:
+        this.throwError(`Invalid project shape: ${projectConfig.projectShape}`);
+        break;
     }
-
-    utilities.log(`${this.testSuiteSuffixName} - ...finished setUpTestingEnvironment()`);
-    utilities.log('');
+    utilities.log(`Project folder to open: ${this.projectFolderPath}`);
+    await utilities.openFolder(this.projectFolderPath!);
```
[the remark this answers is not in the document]
mingxuanzhangsfdx: I am not sure I follow your idea. Is a new project created under NEW "an existing project"?

test/specs/sfdxProjectJson.e2e.ts
```diff
@@ -17,4 +24,8 @@ describe('Customize sfdx-project.json', async () => {
       await utilities.verifyExtensionsAreRunning(utilities.getExtensionsToVerifyActive())
     ).toBe(true);
   });
+
+  after('Tear down and clean up the testing environment', async () => {
+    await testSetup?.tearDown();
```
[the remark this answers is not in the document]
mingxuanzhangsfdx: What if there is something wrong during generating the sf project, so the testSetup object is not returned? (Please refer to the updated implementation of instantiating testSetup as per your suggestion.)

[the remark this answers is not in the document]
peternhale: I'll have a look

test/testSetup.ts
```diff
@@ -297,6 +283,41 @@ export class TestSetup {
...
+      );
+      if (sfDataDeleteRecord.exitCode > 0) {
+        const message = `data delete record failed with exit code ${sfDataDeleteRecord.exitCode}\n stderr ${sfDataDeleteRecord.stderr}`;
+        utilities.log(message);
+        throw new Error(message);
+      }
+    }
+  }
+
+  public updateScratchOrgDefWithEdition(scratchOrgEdition: utilities.OrgEdition) {
```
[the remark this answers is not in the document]
peternhale: That particular test is an edge case. Having this as public will do for now.

test/testSetup.ts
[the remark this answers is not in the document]
peternhale: The switch statement cases have a mix of breaks and returns. What I am suggesting is that they all should have breaks and then resolve the common post switch statement actions. While the code as is works, it is a bit difficult to follow the flow when each case can exit differently.

peternhale (commented):
@mingxuanzhangsfdx looks good. I would like to resolve our discussion in the workspace setup function before merging.

test/specs/auraLsp.e2e.ts
```diff
@@ -14,10 +14,17 @@ const CMD_KEY = process.platform === 'darwin' ? Key.Command : Key.Control;
 describe('Aura LSP', async () => {
   let testSetup: TestSetup;
 
+  const testReqConfig: utilities.TestReqConfig = {
+    projectConfig: {
+      projectShape: utilities.ProjectShapeOption.NEW,
+    },
+    isOrgRequired: false,
+    testSuiteSuffixName: 'auraLsp'
```
daphne-sfdc, line 22:
```suggestion
    testSuiteSuffixName: 'AuraLsp'
```

test/specs/sfdxProjectJson.e2e.ts
```diff
@@ -2,12 +2,19 @@ import { step } from 'mocha-steps';
 import { TestSetup } from '../testSetup.ts';
 import * as utilities from '../utilities/index.ts';
 
+
```
daphne-sfdc, line 5:
```suggestion
```

test/selfSpecs/useExistingProject.test.ts
```diff
@@ -1,20 +0,0 @@
...
- */
-
-import { step } from 'mocha-steps';
-import { TestSetup } from '../testSetup.ts';
-import * as utilities from '../utilities/index.ts';
-
-describe('Use existing project', async () => {
-  let testSetup: TestSetup;
-  step('verify existing project is open', async () => {
-    testSetup = new TestSetup('UseExisProject');
```
CristiCanizales, line 15: Isn't this the old way? Is it supposed to be like that?

test/utilities/commandPrompt.ts
```diff
@@ -119,7 +119,7 @@ export async function waitForQuickPick(
       interval: 500, // Check every 500 ms
       timeoutMsg:
         options.msg ??
-        `Expected to find option ${pickListItem} before ${options.timeout} milliseconds`
+        `Expected to find option ${pickListItem} before ${options.timeout}`
```
daphne-sfdc, line 122: Why is the word "milliseconds" removed here?

test/utilities/lwcUtils.ts
```diff
@@ -32,7 +33,10 @@ export async function createLwc(name: string): Promise<void> {
   log('createLwc() - Select the default directory');
   // Select the default directory (press Enter/Return).
   await inputBox.confirm();
-  await pause(Duration.seconds(3));
+  await checkFileOpen(workbench,
+    name + '.js',
+    { timeout: Duration.seconds(5) }
+  )
```
daphne-sfdc, line 39: Great new validation!  Just a semicolon nit here 🙂 

```suggestion
  await checkFileOpen(workbench,
    name + '.js',
    { timeout: Duration.seconds(5) }
  );
```

test/utilities/commandPrompt.ts
mingxuanzhangsfdx: Because when ${options.timeout} is printed, it comes together with a unit of time such as seconds.

test/selfSpecs/useExistingProject.test.ts
daphne-sfdc, line 1: Just want to confirm - this file is deleted because `metadataDeployRetrieve.e2e.ts` uses the NAMED configuration to open an existing project, correct?

test/selfSpecs/useExistingProject.test.ts
mingxuanzhangsfdx: This was deleted by me. Since I do not think we have used it anywhere.

test/selfSpecs/useExistingProject.test.ts
mingxuanzhangsfdx: Yeah Pete implemented it with the old design. So this test is outdated now and we have already had tests with NAMED already.

test/selfSpecs/useExistingProject.test.ts
CristiCanizales: oh I missed that haha, sorry!

test/specs/apexReplayDebugger.e2e.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, salesforce.com, inc.
+ * Copyright (c) 2024, salesforce.com, inc.
```
daphne-sfdc, line 2: This test was created last year, so 2023 was correct.

CristiCanizales (approved):
great work! Thanks for doing this

test/testSetup.ts
```diff
@@ -1,300 +1,165 @@
 /*
- * Copyright (c) 2023, salesforce.com, inc.
+ * Copyright (c) 2024, salesforce.com, inc.
```
daphne-sfdc, line 2: Was this intentional because you pretty much rewrote the entire testSetup file in this refactor?

test/testSetup.ts
```diff
@@ -1,300 +1,165 @@
...
+          }
+          const repoName = utilities.getRepoNameFromUrl(projectConfig.githubRepoUrl);
+          if (!repoName) {
+            this.throwError(`Unable to determine repository name from URL: ${projectConfig.githubRepoUrl}`);
+          } else {
+            projectName = repoName;
+            if (projectConfig.folderPath) {
+              const localProjName = utilities.getFolderName(projectConfig.folderPath);
+              if (localProjName !== repoName) {
+                this.throwError(`The local project ${localProjName} does not match the required Github repo ${repoName}`);
```
daphne-sfdc, line 96: I assume this is the case where the directory exists, but the contents of the directory is not the desired Salesforce project, right?  For example, if you want to use the dreamhouse-lwc project for your E2E test, and I have a folder called `dreamhouse-lwc` but it is an empty folder?

test/testSetup.ts
mingxuanzhangsfdx: yes

daphne-sfdc (approved):
Thanks Mingxuan for the great refactoring! 🎉

test/testSetup.ts
mingxuanzhangsfdx: This is for the case where we have the project available locally already. So there is no need to clone the repo. It is pretty much making convenience for local testing.

test/specs/apexReplayDebugger.e2e.ts
mingxuanzhangsfdx: Oh yeah, I was testing the header feature against the file so updated this. I think it does not hurt though. Thanks for catching this.

test/testSetup.ts
daphne-sfdc: Yes, I know it's the case where the project is available locally, and we don't need to reclone.  I meant, the if statement is for the rainy day case where the local folder doesn't match the project, right?  So a folder with the project name exists, but it's not the actual project?

PR #104, opened 2024-09-06
Make LWC tests be able to run on GHA

CristiCanizales:
Fixes for LWC tests on GHA.

@W-16475385@

[Test run](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/10731749746/job/29762350040) ✅


test/specs/debugLwcTests.e2e.ts
CristiCanizales, line 1: The only changes in this file are removing the condition `if (process.platform === 'darwin')` and removing on space between the word PASS and the test path (`PASS  path/to/test` --> `PASS path/to/test`)

test/specs/runLwcTests.e2e.ts
CristiCanizales, line 1: The only changes in this file are removing the condition `if (process.platform === 'darwin')` and removing on space between the word PASS and the test path (`PASS  path/to/test` --> `PASS path/to/test`)

test/specs/runLwcTests.e2e.ts
```diff
@@ -12,554 +12,550 @@ import path from 'path';
...
+  });
+
+  step('Run Single Test via the Test Sidebar', async () => {
+    utilities.log(`${testSetup.testSuiteSuffixName} - Run Single Test via the Test Sidebar`);
+    const workbench = await utilities.getWorkbench();
+    const testingView = await workbench.getActivityBar().getViewControl('Testing');
+
+    // Open the Test Sidebar
+    const testingSideBarView = await testingView?.openView();
+    await expect(testingSideBarView).toBeInstanceOf(SideBarView);
```
peternhale, line 297: Seems that if testingView is undefined, the rest of the step is moot.
With not undefined expect the step will fail for the correct reason if the testing activity is not present.
``` suggestion
    const testingView = await workbench.getActivityBar().getViewControl('Testing');
    await expect(testingView).not.toBeUndefined();

    // Open the Test Sidebar
    const testingSideBarView = await testingView?.openView();
    await expect(testingSideBarView).toBeInstanceOf(SideBarView);
```

test/specs/runLwcTests.e2e.ts
```diff
@@ -12,554 +12,550 @@ import path from 'path';
...
+
+    // Hover a test name under one of the test lwc sections and click the run button that is shown to the right of the test name on the Test sidebar
+    const lwcTestsSection = await utilities.getTestsSection(workbench, 'LWC TESTS');
+    const lwcTestItem = (await lwcTestsSection.findItem('displays greeting')) as TreeItem;
```
peternhale, line 301: Please find opportunities to refactor with goal of providing a test framework. I understand the need for the cast given findItem returns a ViewItem, which is the base class of TreeItem. There are a number of places where our extensions use a tree, apex/lwc tests and org browser.

test/environmentSettings.ts
```diff
@@ -48,7 +50,7 @@ export class EnvironmentSettings {
   private _throttleFactor = 1;
   private _javaHome = process.env.JAVA_HOME;
   private _useExistingProject: string | undefined;
-  private _logLevel: LogLevel = 'info';
+  private _logLevel: LogLevel = 'debug';
```
peternhale, line 53: Given the change in wdio.config.ts, it seems this should be `warn`

test/utilities/miscellaneous.ts
```diff
@@ -26,7 +26,7 @@ export function log(message: string): void {
 }
 
 export function debug(message: string): void {
-  if (EnvironmentSettings.getInstance().logLevel in ['debug', 'trace']) {
+  if (EnvironmentSettings.getInstance().logLevel in ['debug', 'trace', 'warn']) {
```
peternhale, line 29: To be in line with accepted understanding of log level, this should include `info` as well.

test/wdio.conf.ts
```diff
@@ -162,7 +162,7 @@ export const config: Options.Testrunner = {
   // Define all options that are relevant for the WebdriverIO instance here
   //
   // Level of logging verbosity: trace | debug | info | warn | error | silent
-  logLevel: EnvironmentSettings.getInstance().logLevel,
+  logLevel: 'warn',
```
peternhale, line 165: This still needs to be configurable. However, given this change, `warn` gives us an acceptable level of console output for typical use. Maybe the value here should be based on the selected log level, via env var, but adjusted to "fit" wdio's application of log levels. If log level is `info` from env, it is then mapped to `warn`, `debug` -> `info`, etc.

peternhale (changes requested):

test/environmentSettings.ts
CristiCanizales: I have an item in the PL to talk about this because I wasn't sure about some stuff about loglevel

peternhale (approved):

test/specs/runLwcTests.e2e.ts
CristiCanizales: Will work on this in follow up PR

PR #109, opened 2024-09-24
fix: validate sobjects related folders

CristiCanizales:
- Adds logging in each step and right validation for presence of folders for sObjectsDefinitions.e2e.ts file

@W-16825635@

[Test run](https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/11022085552) ✅

### DIFF
--- a/test/specs/sObjectsDefinitions.e2e.ts
+++ b/test/specs/sObjectsDefinitions.e2e.ts
@@ -96,7 +96,7 @@ describe('SObjects Definitions', async () => {
     );
     await refreshSObjectDefinitions('Custom SObjects');
 
-    await verifyOutputPanelText(2, 'Custom sObjects');
+    await verifyOutputPanelText('Custom sObjects', 2);
 
     const workbench = await utilities.getWorkbench();
     const treeViewSection = await verifySObjectFolders(workbench, projectName, 'customObjects');
@@ -114,7 +114,7 @@ describe('SObjects Definitions', async () => {
     );
     await refreshSObjectDefinitions('Standard SObjects');
 
-    await verifyOutputPanelText(656, 'Standard sObjects');
+    await verifyOutputPanelText('Standard sObjects');
 
     const workbench = await utilities.getWorkbench();
     const treeViewSection = await verifySObjectFolders(workbench, projectName, 'standardObjects');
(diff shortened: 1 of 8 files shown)


test/specs/sObjectsDefinitions.e2e.ts
```diff
@@ -23,6 +23,9 @@ describe('SObjects Definitions', async () => {
   });
 
   step('Check Custom Objects Customer and Product are within objects folder', async () => {
+    utilities.log(
+      `${testSetup.testSuiteSuffixName} - Check Custom Objects Customer and Product are within objects folder`
```
diyer, line 27: nit: can we change this to "Check Custom Objects 'Customer__c' and 'Product__c' are within objects folder"

diyer: nm, looks like this is an existing step

test/specs/sObjectsDefinitions.e2e.ts
```diff
@@ -86,6 +90,9 @@ describe('SObjects Definitions', async () => {
   });
 
   step('Refresh SObject Definitions for Custom SObjects', async () => {
```
diyer, line 92: not related to this PR, but looks like we can refactor out the step method in this test as it seems to be doing the same thing.

test/specs/sObjectsDefinitions.e2e.ts
```diff
@@ -203,11 +214,34 @@ describe('SObjects Definitions', async () => {
...
+    sobjectsTreeItem = sobjectsTreeItem!;
+
+    await sobjectsTreeItem.expand();
+    await expect(await sobjectsTreeItem.isExpanded()).toBe(true);
+    await utilities.pause(utilities.Duration.seconds(1));
+
+    // Verify if 'standardObjects' folder is within 'sobjects'
+    let standardObjectsTreeItem = await utilities.getVisibleChild(sfdxTreeItem, 'standardObjects');
     await expect(standardObjectsTreeItem).not.toEqual(undefined);
+    standardObjectsTreeItem = standardObjectsTreeItem!;
```
peternhale, line 243: I understand that the expect does not provide TS with enough informaion to continue without applying the `!`, but use of `!` is discouraged. 
```suggestion
    if (!standardObjectsTreeItem) {
      fail('Could not locate standardObjects in tree view');
    }
    ...
```

test/specs/sObjectsDefinitions.e2e.ts
peternhale: There are three paths to check All, Standard and Custom and given we do not have unit tests how else do we confirm that the command is setup correctly?

peternhale (changes requested):
Just a nit on the use of `!` operator. This test has a ton of duplicate code. Would you have a go at that as part of this PR?

peternhale (approved):

PR #111, opened 2024-09-27
Fix flapper in sfdx-project.json caused by the command Go To File...

mingxuanzhangsfdx:
@W-16856300@

sfdx-project.json is not able to be found through getTextEditor. Since getTextEditor uses the command Go To File... which requires VSCode to index the whole project so the indexing time is variable.


CristiCanizales:
Test run: https://github.com/forcedotcom/salesforcedx-vscode/actions/runs/11077692388

CristiCanizales (approved):
LGTM!

